PR #8, opened 2024-06-01
MIMIC example fix + minor improvements

prenc:


MIMIC-IV_Example/joint_script.sh
mmcdermott, line 1: See the changes in #9 -- `set -e` at the top of the script fixes the exit status thing

Issue #30, opened 2024-07-10
MIMIC IV Example has high memory usage due to `*.csv.gz` files


Oufattole:
On my server it uses over 64GBs on the `shard_events.py` script in the MIMIC IV Example:
```bash
echo "Running shard_events.py with $N_PARALLEL_WORKERS workers in parallel"
./scripts/extraction/shard_events.py \
    --multirun \
    worker="range(0,$N_PARALLEL_WORKERS)" \
    hydra/launcher=joblib \
    input_dir="$MIMICIV_PREMEDS_DIR" \
    cohort_dir="$MIMICIV_MEDS_DIR" \
    event_conversion_config_fp=./MIMIC-IV_Example/configs/event_configs.yaml "$@"
```

Memory usage can be significantly reduced by unzipping the files prior to running `shard_events.py` like so:
```bash
echo "Unzipping the files."
for file in ${MIMICIV_PREMEDS_DIR}/*/*.csv.gz; do gzip -d --force "$file"; done
```

PR #31, opened 2024-07-10
added unzipping functionality for mimic iv example

Oufattole:
Resolves issue #30 

## Summary by CodeRabbit

- **New Features**
  - Introduced an optional parameter to control the unzipping of `.csv.gz` files, enhancing user flexibility in data handling based on their storage preferences.
  - Expanded the terminology documentation to provide comprehensive definitions related to MEDS concepts, improving clarity and utility for users.


mmcdermott:
Can you make this flag controlled, so that in the joint script someone can have this not be run if desired? Purely as a bash script thing, I mean, just with an optional 5th arg that must be "do_unzip=true" or "do_unzip=false" or something? This will cost a lot of disk, so some folks may not want to do it, is all.

mmcdermott:
@Oufattole can you merge the changes in main to this, make it go into dev instead of main, and make this flag controlled via the shell parameters? Then I'm happy to merge it.

MIMIC-IV_Example/joint_script.sh
```diff
@@ -42,6 +42,9 @@ shift 4
 echo "Running pre-MEDS conversion."
 ./MIMIC-IV_Example/pre_MEDS.py raw_cohort_dir="$MIMICIV_RAW_DIR" output_dir="$MIMICIV_PREMEDS_DIR"
 
+echo "Unzipping the pre-MEDS gunzip csv files."
+for file in ${MIMICIV_PREMEDS_DIR}/*/*.csv.gz; do gzip -d --force "$file"; done
```
mmcdermott, line 46: actually, you probably want to do this on the raw data, right? That way the pre-meds can benefit from faster processing too?

Oufattole: Done!

mmcdermott:
@Oufattole the test failure here is due to a polars versioning issue. I'm going to fix it directly then you can merge that change in and we'll push this.

MIMIC-IV_Example/joint_script.sh
```diff
@@ -36,9 +36,17 @@ MIMICIV_RAW_DIR="$1"
 MIMICIV_PREMEDS_DIR="$2"
 MIMICIV_MEDS_DIR="$3"
 N_PARALLEL_WORKERS="$4"
+DO_UNZIP="${5:-do_unzip=false}"  # Default to 'do_unzip=false' if not provided, this uses more disk space but reduces memory usage
```
mmcdermott, line 39: You need to change one more thing here -- this script is set up so that the first 4 args are mandatory, and all others are sent to the scripts as arguments via the `"$@"` thing. Right now, your `DO_UNZIP` will absorb all of those other kwargs. You should either:

1. Make sure that `DO_UNZIP` only absorbs the fifth argument if the arg there starts with `--do_unzip` or something.
- or -
2. Make there be a mandatory 5th arg that is `DO_UNZIP` and then replace the `shift 4` with `shift 5` so that it isn't used again below in the `"$@"`

mmcdermott:
@Oufattole I fixed the test issue but added one comment on your param and also you should update the display help function for this param.

mmcdermott:
@Oufattole, @EthanSteinberg discovered that this won't work even as implemented here b/c `pre_MEDS.py` only looks for `csv.gz` files, not `csv` files. I'm going to debug and try to fix and push this in today.

mmcdermott:
Re-opened with changes to `pre_MEDS.py` and some other changes which seems to be working in that unzipping can occur and the extraction pipeline proceeds through the `shard_events` stage with the unzipped files. I haven't profiled the impact this has on memory but I believe it should be non-trivial.

PR #46, opened 2024-07-19
Maintain the order of event from event_config while merging cohorts

prenc:
## Summary by CodeRabbit

- **New Features**
	- Enhanced file reading functionality that allows for targeted file path construction based on specific event subsets.
	- Improved error messages to provide clearer feedback on file retrieval issues.
	- Introduced order preservation in the uniqueness of values processed, ensuring data integrity during operations.

- **Bug Fixes**
	- Fixed error handling to ensure consistency in feedback regarding file not found scenarios.
	- Added checks to ensure the number of directories matches the provided event subsets, enhancing robustness.


mmcdermott:
so this makes it so that if your event conversion config is like this one https://github.com/mmcdermott/MEDS_polars_functions/blob/main/MIMIC-IV_Example/configs/event_configs.yaml it will do all the `hosp/admission` events before the `hosp/diagnoses_icd` before the `hosp/drgcodes`, etc. But, if I'm reading this right, it won't necessarily make it so that within `hosp/admissions`, `ed_registrations` will come before `ed_out` will come before `admission`, right? Is that a problem?

What you really want is to be able to assert a partial order of codes or code prefixes, and have the data be re-ordered within the patient ID and timestamp ordering to respect that partial order, right? Can we add that capability directly via a dedicated transformation?

mmcdermott:
What about something like https://github.com/mmcdermott/MEDS_polars_functions/pull/48 instead? That way this is functionality you can apply after the extraction process too on any MEDS dataset, regardless of the source of the event conversion configs file?

scripts/extraction/merge_to_MEDS_cohort.py
```diff
@@ -32,19 +37,19 @@ def read_fn(sp_dir: Path, unique_by: list[str] | str | None) -> pl.LazyFrame:
...
                     subset.append(u)
                 else:
                     logger.warning(f"Column {u} not found in dataframe. Omitting from unique-by subset.")
-            df = df.unique(maintain_order=False, subset=subset)
+            df = df.unique(maintain_order=True, subset=subset)
         case _:
             raise ValueError(f"Invalid unique_by value: {unique_by}")
 
-    return df.sort(by=["patient_id", "timestamp"], multithreaded=False)
+    return df.sort(by=["patient_id", "timestamp"], maintain_order=True)
```
mmcdermott, line 52: Keep the `multithreaded=False` please or make it configurable via the stage config. It can get expensive for datasets with lots of measurements, and eICU in particular was giving me trouble.

PR #48, opened 2024-07-19
Added `reorder_measurements` transformation.

### LINKED ISSUE #47: Add transformation for re-ordering measurements

mmcdermott:
This allows you to specify a list of regexs in a config variable and measurements will be sorted, within a patient-id X timestamp level, by the earliest index of the regex they satisfy.  This would close #47 

Still needed:
- [x] Updates to documentation
- [x] An added integration test similar to the other transform tests in https://github.com/mmcdermott/MEDS_transforms/tree/main/tests
- [x] Probably some fixes to the code itself.
- [x] doctests


mmcdermott:
This should wait for #107 to go in as it will allow us to simplify the interface on this new transformation slightly.

mmcdermott:
@prenc I actually merged this by accident, but I think it should work for your use case and don't want to revert it. Please let me know if you see any issues.

PR #81, opened 2024-07-30
51 add aggregation for quantile computation of code values

Oufattole:
## Summary by CodeRabbit

- **New Features**
  - Introduced aggregation functionality for calculating quantiles of numerical values, enhancing metadata analysis.
  - Added support for a new quantile aggregation type in the metadata processing workflow.

- **Documentation**
  - Expanded documentation for key functions to include details and examples for the new quantile aggregation feature. 

- **Chores**
  - Updated function signatures to enhance flexibility for testing scenarios.

### DIFF
--- a/src/MEDS_transforms/aggregate_code_metadata.py
+++ b/src/MEDS_transforms/aggregate_code_metadata.py
@@ -62,7 +62,8 @@ class METADATA_FN(StrEnum):
             the code
         "values/min": Collects the minimum non-null, non-nan numerical_value value for the code & modifiers
         "values/max": Collects the maximum non-null, non-nan numerical_value value for the code & modifiers
-        "values/quantiles": TODO
+        "values/quantiles": Collects the quantiles of the non-null, non-nan numerical_value value for the
+            code & modifiers
     """
 
     CODE_N_PATIENTS = "code/n_patients"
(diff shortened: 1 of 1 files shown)


src/MEDS_transforms/aggregate_code_metadata.py
```diff
@@ -204,7 +230,9 @@ def validate_args_and_get_code_cols(
 
 
 def mapper_fntr(
-    stage_cfg: DictConfig, code_modifier_columns: list[str] | None
+    stage_cfg: DictConfig,
+    code_modifier_columns: list[str] | None,
+    test=False,
```
mmcdermott, line 235: Does this parameter still belong? I think likely not, correct?

src/MEDS_transforms/aggregate_code_metadata.py
```diff
@@ -99,6 +102,25 @@ class MapReducePair(NamedTuple):
     reducer: Callable[[pl.Expr | Sequence[pl.Expr] | cs._selector_proxy_], pl.Expr]
 
 
+def quantile_reducer(df: pl.LazyFrame) -> pl.Expr:
+    agg = "values/quantiles"
+    cols = cs.matches(f"{agg}/shard_\\d+")
+    quantile_df = df.select(~cols, pl.concat_list(cols.fill_null([])).alias(agg)).explode(agg).unnest(agg)
+    quantile_df = (
+        quantile_df.with_columns(pl.col("numerical_value").repeat_by("count"))
```
mmcdermott, line 110: If you're doing it this way, then just keep things computationally simple and store the values with repetition and don't worry about the counts. If you want to store the values and their occurrence counts, then use those counts in a smarter way to determine the quantile (e.g., take a cumsum of the counts when the values are sorted by value then find the position in that list corresponding to the quantile of interest).

src/MEDS_transforms/aggregate_code_metadata.py
```diff
@@ -99,6 +102,25 @@ class MapReducePair(NamedTuple):
...
+        .explode("numerical_value")
+        .drop("count")
+    )
+    quantile_keys = [0.25, 0.5, 0.75]
```
mmcdermott, line 114: We need to find a way to parametrize this. That is the real challenge here, I think, from an interface perspective. These three random quantiles aren't what people will actually want.

src/MEDS_transforms/aggregate_code_metadata.py
```diff
@@ -394,155 +437,197 @@ def mapper(df: pl.LazyFrame) -> pl.LazyFrame:
...
+            code_modifier_columns: A list of column names that should be used in addition to the core `code`
+                column to group the data before applying the aggregations. If None, only the `code` column
+                will be used.
+
+        Returns:
+            A function that aggregates the specified metadata columns from different extracted metadata shards
+            into a total view.
+
+        Raises: See `validate_args_and_get_code_cols`.
+    [
```
mmcdermott, line 458: This doesn't look like something you meant to keep around?

mmcdermott (commented):
Did you intend for this to be ready for review? Or did I misunderstand something? I can finish a full, detailed review if you did intend it to be ready, but I wasn't sure given some of the parts if I might be being premature.

mmcdermott:
Take a look at this approach for the simplicity of just keeping the values free: https://github.com/mmcdermott/MEDS_transforms/compare/dev...alt_51_raw_values?expand=1

It doesn't solve the interface problem though.

mmcdermott:
You might do something like this for parametrization, so you could pass `"values/quantiles/0.1,0.2,0.3" in or something:

https://github.com/mmcdermott/MEDS_transforms/compare/alt_51_raw_values...starter_code_for_params?expand=1

Just a starting thought.

mmcdermott:
@Oufattole I suggest we close this in favor of #98 for now. Do you have any objections?

mmcdermott:
Closing in favor of #98

Issue #51, opened 2024-07-21
Add aggregation for quantile computation of code values


mmcdermott:
This should go into the [`aggregate_code_metadata.py`](https://github.com/mmcdermott/MEDS_polars_functions/blob/dev/src/MEDS_polars_functions/aggregate_code_metadata.py) file, which means we need
  - [ ] A way to define the aggregation such that it encompasses the number of quantiles being requested that can be used in the [`aggregate_code_metadata.py`](https://github.com/mmcdermott/MEDS_polars_functions/blob/dev/src/MEDS_polars_functions/aggregate_code_metadata.py) file (which will require some requisite changes potentially).
  - [ ] a mapper expression that computes some statistics per shard capable of being used to reconstitute approximate quantiles,
  - [ ] A reducer function that aggregates those statistics into said quantiles
  - [ ] A storage format for compound outputs after mapreduce (e.g., are quantiles stored as a list? An array? A struct?)
  
There are a few options for each of these.

## Parameters
For the parametrization question, the easiest approach is just to expect a dedicated config parameter (e.g., `n_quantiles`) that is outside the list of aggregations of interest. This means you could only compute one set of quantiles per aggregation stage, but makes things trivial to incorporate into the existing [`aggregate_code_metadata.py`](https://github.com/mmcdermott/MEDS_polars_functions/blob/dev/src/MEDS_polars_functions/aggregate_code_metadata.py) structure. This is likely the best starting point. The right long term solution is to make the aggregation names be "parameterized constants" instead of pure constants, so that you can do something like `"values/quantiles/5"` or `"values/quantiles/100@0.3"` or something instead to specify the number of quantiles and (optionally) the error parameter for approximation algorithms (said algorithms are needed as you can't easily compute exact quantiles across a sharded dataframe)

## Mapping and Reducing
To do the mapreduce properly, we can either include an approximate quantile algorithm, such as by polarsifying this [ddsketch](https://github.com/DataDog/sketches-py/blob/master/ddsketch/ddsketch.py) algorithm. Alternatively, we can do something more memory intensive but also more accurate that is likely sufficient and not too expensive for our needs provided we can restrict the precision of the numerical values observed to a certain # of digits which is to just compute the full count of each unique value observed or otherwise compute a density approximation or histogram approximation and use that to compute the quantiles after the reduction step is finalized. This would make the mapping and reduction easier, but more expensive computationally, but also would require some post-reduction step.

## Storage format
[...]

PR #98, opened 2024-08-02
Support quantile computation over numerical values in a simple (though inefficient) manner.

mmcdermott:
I think this should work for #51 for now in place of #81


Oufattole (approved):

prenc (approved):
Looks awesome to me, this is what I need

PR #119, opened 2024-08-05
Adds the ability to specialize existing transformations to be applied with separate configurations to different portions of the data based on configurable filters.

### LINKED ISSUE #101: Add match & revise syntax transformations
Suggested syntax / configuration.

Suppose a transformation has a `stage_cfg` that normally takes some arguments. As all `stage_cfg`s are dictionaries, we can do the following:

If (for transformations that support match & revise), the `stage_cfg` has a `_match_and_revise` key, then:
  1. All other `stage_cfg` keys are assumed to be global default values.
  2. An error is thrown if each element of the `_match_and_revise` is not a list of dictionaries with each having at least a `_matcher` key which points to a valid `match_configuration` object (validity criteria yet to be determined).
  3. The transformation will be run for each element of the `_match_and_revise` list filtered to only those measurements that match the `_matcher` key value, using all other keyword arguments of the list element as the `stage_cfg` arguments for that revision run, defaulting to global arguments.

mmcdermott:
Adds a sort of "meta" stage configuration set-up in which transforms can be made to be applied to portions of dataframes conditionally via specified filters. This paradigm is called the "match & revise" pattern. Once merged, this will close #101


mmcdermott:
@prenc I know this is still quite preliminary, but does this look like it would capture what you're wanting out of match & revise?

PR #121, opened 2024-08-06
Adds an 'extract_values' transform to extract values and retype them from input MEDS data.

### LINKED ISSUE #57: Add `extract_numeric_values` transformation
This may leverage the match & revise syntax of #101

mmcdermott:
This relies on the `parser.py` file for the extraction DSL, which will need to be expanded to support regexes before this becomes truly useful. This will also rely on the match and revise logic (#119) to be set such that values can be extracted in a consistent manner. 

This would close #57 

Needed:
- [x] More extensive doctests
- [x] Integration test
- [x] Likely change `categoric_value` to `categorical_value`
- [x] Regex support


mmcdermott:
@prenc I think this test https://github.com/mmcdermott/MEDS_transforms/pull/121/files#diff-90ce2c41e152f3829a020556def04b9b1a7104dbd263dd891aa52834c668b89e shows the kind of usage you'd want, right?

PR #124, opened 2024-08-08
Fix split_and_shard_patients when the full split definition is provided

prenc:
Fixes two small issues:
 - when the full split definition provided, it resulted in an empty split error
 - wrong cfg.split_json_fp access

## Summary by CodeRabbit

- **New Features**
	- Introduced quantile aggregation capabilities in the metadata processing system, allowing for dynamic specification of quantiles.
	- Automated generation of code reference pages for documentation, enhancing the user experience.

- **Improvements**
	- Enhanced logic for handling patient splits, improving clarity and efficiency.
	- Updated logging for external split configurations to reflect new access paths.
	- Improved readability through the use of streamlined dictionary comprehensions.
	- Added logging to handle cases with no patients to split, preventing potential errors.

- **Documentation**
	- Improved formatting and readability of various documentation files, including README and pipeline configuration documents.
	- Added a new configuration file for Read the Docs, facilitating better documentation management.
	- Established a comprehensive MkDocs configuration for the project documentation.


src/MEDS_transforms/extract/split_and_shard_patients.py
```diff
@@ -231,7 +236,7 @@ def main(cfg: DictConfig):
         if not external_splits_json_fp.exists():
             raise FileNotFoundError(f"External splits JSON file not found at {external_splits_json_fp}")
 
-        logger.info(f"Reading external splits from {cfg.external_splits_json_fp}")
+        logger.info(f"Reading external splits from {cfg.stage_cfg.external_splits_json_fp}")
```
mmcdermott, line 239: `logger.info(f"Reading external splits from {str(external_splits_json_fp.resolve())}")`

src/MEDS_transforms/extract/split_and_shard_patients.py
```diff
@@ -99,31 +99,36 @@ def shard_patients[
...
-    patients = rng.permutation(patient_ids_to_split)
-    patients_per_split = np.split(patients, split_lens.cumsum())
+        patients = rng.permutation(patient_ids_to_split)
+        patients_per_split = np.split(patients, split_lens.cumsum())
 
-    splits = {k: v for k, v in zip(split_names, patients_per_split)}
-    splits = {**splits, **external_splits}
+        splits = {**{k: v for k, v in zip(split_names, patients_per_split)}, **splits}
+    else:
+        logger.info(
```
mmcdermott, line 128: In the event that this clause is triggered _and_ `split_fracs_dict` is not empty, we should either throw an error or, at the very least, use `logger.warning` to catch the fact that none of the splits they may have expected to create will be created. When `split_fracs_dict` is also empty, then just doing `logger.info` is fine.

mmcdermott (approved):
Can you add a doctest that would fail on the old code but passes with your changes so we know if this functionality regresses in the future? After addressing this and the other small comments this is ready to go in.

src/MEDS_transforms/extract/split_and_shard_patients.py
prenc: done!

src/MEDS_transforms/extract/split_and_shard_patients.py
prenc: yes, you're right! it's important

src/MEDS_transforms/extract/split_and_shard_patients.py
```diff
@@ -76,6 +76,16 @@ def shard_patients[
...
+        >>> external_splits = {
+        ...     'train': np.array([1, 2, 3, 4, 5, 6], dtype=int),
+        ...     'test': np.array([7, 8, 9, 10], dtype=int),
+        ... }
+        >>> shard_patients(patients, 6, external_splits)
+        {'train/0': [1, 2, 3, 4, 5, 6], 'test/0': [7, 8, 9, 10]}
+        >>> shard_patients(patients, 3, external_splits)
+        Traceback (most recent call last):
+            ...
+        ValueError: External splits must have fewer patients than n_patients_per_shard (3): len(train)=6, ...
```
mmcdermott, line 88: You don't want it to just re-shard the external splits? Why can't it re-shard the external splits?

prenc: You're right. It should be able to reshard them in this case. So far, it's been breaking on this, I will fix that.

mmcdermott (approved):

PR #143, opened 2024-08-11
Fixes brittle reduce stage checking in Reshard stage.

mmcdermott:
Makes file existence checkers wait for valid parquet files, not just existing files.


mmcdermott:
The issue is not a hydra issue at all; I need to rwlock the shards map computation too.

PR #251, opened 2025-02-17
Handle Datetime type for time column during extraction

prenc:
If the input files are in Parquet format, their time columns may already be of type Datetime. In such cases, we should not parse them as strings but instead use them as they are.

In addition, if the type is Date we could cast it to Datetime, let me know if it also should be added.


mmcdermott (approved):
LGTM -- thanks @prenc! You can also solve this by just omitting the `time_format` key in the config, but this is more robust so is a definite improvement!

PR #252, opened 2025-02-17
Allow ignoring split fracs that are null

prenc:
Added a check to filter out splits with `None` (null in yaml) as their fraction and log a message indicating which splits are being ignored.


mmcdermott (approved):
LGTM -- thanks @prenc, and sorry I haven't had a chance to look into this further yet!

PR #265, opened 2025-03-07
263 drop table flag

### LINKED ISSUE #263: Add flag that allows us to to drop tables that are in the event config
For some datasets/standards, we want to be able to exclude tables that are not processed, not part of the cohort, or cause information leakage.

rvandewater:
Closes #263 I am not that happy with how I had to add code in multiple places, but the problem is that the event_config is read in multiple places such that we cannot remove it once. Happy to hear of a better alternative @mmcdermott 

## Summary by CodeRabbit

- **Chores**
  - Updated dependency configuration to streamline requirements and enhance testing options.

- **Refactor**
  - Streamlined the data preprocessing workflow by removing redundant transformation stages.
  - Improved configuration handling to better manage ignored tables and enhanced logging and error messages during data merging.

- **Tests**
  - Removed outdated tests related to the removed transformation stages and updated test configurations accordingly.

### DIFF
--- a/src/MEDS_transforms/extract/convert_to_sharded_events.py
+++ b/src/MEDS_transforms/extract/convert_to_sharded_events.py
@@ -851,7 +851,11 @@ def main(cfg: DictConfig):
     event_conversion_cfg = OmegaConf.load(event_conversion_cfg_fp)
     logger.info(f"Event conversion config:\n{OmegaConf.to_yaml(event_conversion_cfg)}")
 
-    event_conversion_cfg.pop("tables_to_ignore", [])
+    tables_to_ignore = event_conversion_cfg.pop("tables_to_ignore", None)
+    if tables_to_ignore:
+        logger.warning(f"Ignoring tables: {tables_to_ignore}")
+        for table in tables_to_ignore:
+            event_conversion_cfg.pop(table, None)
     default_subject_id_col = event_conversion_cfg.pop("subject_id_col", "subject_id")
 
     subject_subsharded_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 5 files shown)


src/MEDS_transforms/extract/convert_to_sharded_events.py
```diff
@@ -868,6 +873,7 @@ def main(cfg: DictConfig):
     for sp, subjects in subject_splits:
         for input_prefix, event_cfgs in event_configs:
             event_cfgs = copy.deepcopy(event_cfgs)
+            logger.info(event_cfgs)
```
mmcdermott, line 876: Remove

src/MEDS_transforms/extract/shard_events.py
```diff
@@ -227,7 +227,14 @@ def retrieve_columns(event_conversion_cfg: DictConfig) -> dict[str, list[str]]:
     prefix_to_columns = {}
 
     default_subject_id_col = event_conversion_cfg.pop("subject_id_col", subject_id_field)
+    tables_to_ignore = event_conversion_cfg.pop("tables_to_ignore", None)
+    if tables_to_ignore:
+        logger.warning(f"Ignoring tables: {tables_to_ignore}")
```
mmcdermott, line 232: I think this shouldn't be a warning necessarily as these were requested to be ignored.

mmcdermott (commented):
With this style of approach, why not just remove those tables from the event config? I don't think this adds much flexibility to the system. Better would be to add an option to the main extract config rather than the event configs with a list of tables to ignore (as then you can modify it on the command line even while keeping the core config file the same) or even just add a boolean flag that says one should ignore missing tables. You'll still likely need to modify the code in the same several places, but it will I think be better than this approach.

rvandewater:
@mmcdermott , I implemented creating a "temporary"event_configs which drops the unavailable tables automatically in OMOP MEDS ( (will add a flag to do this though) https://github.com/rvandewater/OMOP_MEDS.

PR #267, opened 2025-03-25
fix: remove mutable default argument in shard_subjects

kirilklein:
Replace the mutable dictionary default argument with None to avoid the 
common Python pitfall where default mutable arguments are created once 
at function definition time. Default split fractions are now assigned 
within the function body when split_fracs_dict is None.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the flexibility of data splitting configurations. Now, if custom split fractions aren’t provided, standard values are automatically applied. This change streamlines configuration and continues to enforce robust error checking for invalid split inputs.

### DIFF
--- a/src/MEDS_transforms/extract/split_and_shard_subjects.py
+++ b/src/MEDS_transforms/extract/split_and_shard_subjects.py
@@ -20,7 +20,7 @@ def shard_subjects(
     subjects: np.ndarray,
     n_subjects_per_shard: int = 50000,
     external_splits: dict[str, Sequence[int]] | None = None,
-    split_fracs_dict: dict[str, float] | None = {"train": 0.8, "tuning": 0.1, "held_out": 0.1},
+    split_fracs_dict: dict[str, float] | None = None,
     seed: int = 1,
 ) -> dict[str, list[int]]:
     """Shard a list of subjects, nested within train/tuning/held-out splits.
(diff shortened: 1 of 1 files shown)


src/MEDS_transforms/extract/split_and_shard_subjects.py
```diff
@@ -93,6 +93,8 @@ def shard_subjects(
         >>> shard_subjects(subjects, 3, external_splits)
         {'train/0': [5, 1, 3], 'train/1': [2, 6, 4], 'test/0': [10, 7], 'test/1': [8, 9]}
     """
+    if split_fracs_dict is None:
```
mmcdermott, line 96: With this check you can update the if statement on line 119 as well, as it can no longer be `None`

mmcdermott: I also don't think this is semantically equivalent; the doctest on line 91 seems to expect different behavior.

mmcdermott:
@kirilklein, I think we're going to move forward with https://github.com/mmcdermott/MEDS_transforms/pull/269 which would put the code relevant to this PR in a different package, https://github.com/mmcdermott/MEDS_extract -- though it may take some time to get all of that to work properly. How urgent is this problem? Does this warrant a hotfix into main (once tests are passing)

kirilklein:
> @kirilklein, I think we're going to move forward with https://github.com/mmcdermott/MEDS_transforms/pull/269 which would put the code relevant to this PR in a different package, https://github.com/mmcdermott/MEDS_extract -- though it may take some time to get all of that to work properly. How urgent is this problem? Does this warrant a hotfix into main (once tests are passing)

Ok, I found a workaround. Also, I figured that this change doesn't fix the issue. We can close this PR

Issue #285, opened 2025-03-28
[Proposal] Consider having main entry run point explicitly take both the pipeline config yaml (or package) and the stage name, and register stages without a pipeline.


mmcdermott:
> So, one additional challenge with this is that the intention for the generic stages in the MEDS-Transforms repo (or even in derived packages) is that they could be used in further packages on their own. So, registering a stage in a pipeline specific manner is therefore an issue. This can be solved by moving the final hydra main construction to the actual main entry point, rather than at registration time, so it can be done in a pipeline specific manner, rather than in a stage specific manner. This is likely a general improvement in design, though could be done independently of this bigger change. 

 _Originally posted by @mmcdermott in [#283](https://github.com/mmcdermott/MEDS_transforms/issues/283#issuecomment-2761672143)_

PR #286, opened 2025-03-28
Breaking adjustments to stage configuration to reduce duplication and simplify and future proof inputs.

### LINKED ISSUE #283: [Proposal] Stage configuration should be adjusted for brevity, to take advantage of dictionary ordering, and future proofing.
Concrete proposal is that instead of:

```yaml
...
stage_configs:
  stage_B: {...}
  stage_A: {...}

stages:
  - stage_A
  - stage_B
  - stage_C

...
```

We should have
```yaml
...
stages:
  - stage_A: {...}
  - stage_B: {...}
  - stage_C

...
```

The reason to use a list in stages instead of just a plain object is that, in the future, we may want to have multiple stages combined to run without writing to/from disk. Plus, it is a functionally non-existent complexity cost while making it exceedingly clear that order matters.

mmcdermott:
Still very much in progress.

Blocked by #285 

Closes #283


mmcdermott:
Closed in favor of #308

PR #287, opened 2025-03-28
Change stage registration to register stages without an explicit pipeline and pass the pipeline file on the command line.

### LINKED ISSUE #285: [Proposal] Consider having main entry run point explicitly take both the pipeline config yaml (or package) and the stage name, and register stages without a pipeline.

mmcdermott:
This better aligns with the fact that stages should be usable across many pipelines, not a single one that is known at registration time.

Closes #285

### DIFF
--- a/src/MEDS_transforms/__main__.py
+++ b/src/MEDS_transforms/__main__.py
@@ -1,7 +1,17 @@
+import os
 import sys
 from importlib.metadata import entry_points
+from importlib.resources import files
+from pathlib import Path
+
+import hydra
+from omegaconf import OmegaConf
 
 from . import __package_name__
+
+HELP_STRS = {"--help", "-h", "help", "h"}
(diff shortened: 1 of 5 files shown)


mmcdermott:
The earlier failure I mentioned was a local environment / pycache issue. It does not persist on a clean install.

PR #289, opened 2025-03-28
259 dry testing code further

mmcdermott:


mmcdermott:
This is getting absorbed into #290

Issue #366, opened 2026-04-10
StageExample metadata check_outputs should support allow_extra_columns


mmcdermott:
## Problem

`StageExample.check_outputs()` for metadata stages (lines 1020-1040) compares expected vs actual columns strictly — if the actual output has columns not in the expected metadata, it raises `AssertionError: Column mismatch: want-only=set(), got-only={...}`.

There's no way to opt into allowing extra columns in the actual output.

## Use case

MEDS_extract's `extract_code_metadata` stage now adds a `code_template` column to `metadata/codes.parquet`. Downstream test examples that were written before this column existed specify expected metadata without `code_template`. They need `allow_extra_columns=True` behavior (like what `assert_frame_equal` doesn't natively support, but which MEDS_extract's own `assert_df_equal` helper implements).

## Proposed solution

Add an `allow_extra_columns` option to `df_check_kwargs` (or a separate field in `_test_cfg.yaml`) that, when True, only checks columns present in `want_metadata` rather than requiring exact column match. Something like:

```python
if not self.df_check_kwargs.get("allow_extra_columns", False):
    if want_only or got_only:
        raise AssertionError(...)
elif want_only:
    raise AssertionError(...)  # Missing expected columns is always an error

got_metadata = got_metadata.select(self.want_metadata.columns)
```

## Context

This came up while migrating MEDS_extract tests to the StageExample system. The `extract_code_metadata` stage's test uses `allow_extra_columns=True` because the stage adds provenance columns (`code_template`) that the test's expected output doesn't enumerate.

mmcdermott:
## Assessment: Non-MEDS Output Support via yaml_to_disk

I evaluated adding non-MEDS output support following the same `yaml_to_disk` pattern used for inputs (PRs #362-364). 

**Verdict: Not a minimal change.** The input side was simple because you only *expand* YAML into files at test time without validation. The output side requires *comparing* actual output directories against expected YAML-described structures, which is substantially more involved:

1. **Type hint / `from_dir` changes** — trivial (~10 lines, same pattern as input)
2. **`check_outputs` rewrite** — the hard part (~60-80 lines). Currently assumes MEDS structure (`data/*.parquet`, `metadata/codes.parquet`). For arbitrary outputs you'd need:
   - Expand expected YAML via `yaml_to_disk` into a temp dir
   - Recursive file-by-file comparison against actual output
   - Handle different file formats (CSV, JSON, parquet) with appropriate comparison logic
[...]

PR #368, opened 2026-04-11
Allow StageExample subclassing via example_class parameter

mmcdermott:
## Summary
- Add `example_class` parameter to `Stage.__init__` (defaults to `StageExample`)
- Use `self.example_class` in `Stage.test_cases` instead of hardcoded `StageExample` reference
- Downstream packages can now provide custom `StageExample` subclasses with overridden `check_outputs`, `from_dir`, etc.

Relates to #366

## Docgen impact
**Safe for the common case.** Subclasses that override `check_outputs` but keep `want_data` as `MEDSDataset` and `want_metadata` as `pl.DataFrame` render correctly. If a future subclass changes field types to `Path`, `_format_example()` in `docgen.py` would need `isinstance` guards — but that's a separate change. See discussion on #366.

## Example downstream usage
```python
class FlexibleStageExample(StageExample):
    def check_outputs(self, output_dir, is_resolved_dir=False):
        # Custom validation tolerating extra columns, non-MEDS outputs, etc.
        ...

@Stage.register(example_class=FlexibleStageExample)
def my_custom_stage(...):
    ...
```

## Test plan
- [x] All 91 doctests pass
- [x] Pre-commit hooks pass
- [x] Change is 5 lines across 1 file — purely additive, no behavior change for existing stages


example/simple_example_pkg/src/simple_example_pkg/export_code_summary/export_code_summary.py
```diff
@@ -0,0 +1,153 @@
...
+            ...
+        AssertionError: JSON mismatch in code_summary.json...
+    """
+
+    # Override to accept Path (directory of expected JSON files) instead of MEDSDataset
+    want_data: Path | None = None
+
+    def __post_init__(self):
+        if self.want_data is None and self.want_metadata is None:
+            raise ValueError("Either want_data or want_metadata must be provided.")
```
mmcdermott: This doesn't seem right for this class.

example/simple_example_pkg/src/simple_example_pkg/export_code_summary/export_code_summary.py
```diff
@@ -0,0 +1,153 @@
...
+        if self.df_check_kwargs is None:
+            self.df_check_kwargs = {"rel_tol": 1e-3, "abs_tol": 1e-5}
```
mmcdermott: Ditto

example/simple_example_pkg/src/simple_example_pkg/export_code_summary/export_code_summary.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @classmethod
+    def from_dir(cls, stage_name, scenario_name, example_dir, **schema_updates):
+        """Parse example directory, treating out_data.yaml as a yaml_to_disk path for JSON output."""
+        want_data_fp = example_dir / "out_data.yaml"
+        in_fp = example_dir / "in.yaml"
+
+        want_data = want_data_fp if want_data_fp.is_file() else None
+        in_data = None
+        if in_fp.is_file():
```
mmcdermott: Why do you need this? Why can't you use the base class for this?

example/simple_example_pkg/src/simple_example_pkg/export_code_summary/export_code_summary.py
```diff
@@ -0,0 +1,153 @@
...
+            from MEDS_transforms.stages.examples import MEDSDataset
```
mmcdermott: Is this the right import source for this?

example/simple_example_pkg/src/simple_example_pkg/export_code_summary/examples/README.md
```diff
@@ -0,0 +1,11 @@
+# export_code_summary example
+
+This example demonstrates the `example_class` parameter on `Stage.register`.
+
+The stage writes a JSON file (`code_summary.json`), not MEDS-format parquet. The default
+`StageExample.check_outputs` expects `data/*.parquet` or `metadata/codes.parquet` and would fail
+here. By registering with `example_class=JsonOutputStageExample`, the stage uses a subclass that
+validates output by comparing JSON files via `yaml_to_disk` instead.
+
+The `out_data.yaml` file here is a `yaml_to_disk` specification that describes the expected JSON
```
mmcdermott: Is it actually using yaml_to_disk, though?

mmcdermott:
Addressed the review in 3f38047:

- **`__post_init__` override** (comments on lines 66 and 68): Dropped entirely. The base class's `__post_init__` already validates `want_data`/`want_metadata` correctly and the `df_check_kwargs` default is harmless when unused.
- **Duplicated `in_fp` handling in `from_dir`**: This example has no `in.yaml` (the stage reads MEDS data produced by the pipeline), so the override now just returns `cls(..., want_data=example_dir / "out_data.yaml")` — no duplication of base-class logic.
- **`MEDSDataset` import source**: Removed — no longer needed after the `from_dir` simplification. (The canonical import is `meds_testing_helpers.dataset`, as used in the base class.)
- **README `yaml_to_disk` claim**: Now actually true. `check_outputs` uses `yaml_disk` to materialize the expected spec into a temp directory, then walks the tree and compares each file against the actual output (JSON as parsed objects, others as text). README updated to match.

All local tests pass (simple_example_pkg: 5/5, main repo `examples.py` doctests: 4/4). Monitoring CI now.

PR #374, opened 2026-04-16
Fix mapreduce reducer race condition (closes #373)

### LINKED ISSUE #373: Flaky CI: reducer reads partially-written parquet from mapper (PAR1 ComputeError)
## Summary

CI on `main` intermittently fails in the `Run tests with joblib` step with:

```
polars.exceptions.ComputeError: parquet: File out of specification: The file must end with PAR1
```

This is a **race between the mapper and reducer** in `mapreduce/reducer.py`: the reducer considers a mapper's output "ready" as soon as `fp.is_file()` is true, but `df.write_parquet(...)` creates the file well before flushing the parquet footer. The reducer then reads a half-written file.

Most recent failure: [Tests run 24377604771](https://github.com/mmcdermott/MEDS_transforms/actions/runs/24377604771), Python 3.12 job only (3.11 passed on the same commit). Error emerged in `fit_normalization`, a `mapreduce_stage`.

## Root Cause

`src/MEDS_transforms/mapreduce/reducer.py:206-208`:

```python
while not all(fp.is_file() for fp in in_fps):
    logger.info("Waiting to begin reduction for all files to be written...")
    time.sleep(polling_time)

reduced = reduce_fn(*[read_fn(fp) for fp in in_fps])
```

`fp.is_file()` returns `True` the instant the mapper's `df.write_parquet(...)` creates the file, even if no bytes — or only the parquet header — have been flushed yet. The reducer then calls `read_fn(fp)` on an incomplete file.

The existing `rwlock_wrap` correctly serializes *writers* (via `filelock.FileLock`), but the reducer is a separate process that does not participate in the lock — it only polls for file existence.

## Reproduction (≈100% reliable)

```python
import tempfile, threading, time
from pathlib import Path
import polars as pl
from MEDS_transforms.mapreduce.reducer import reduce_over

def slow_write(df, fp):
    fp.touch()          # file exists but empty
    time.sleep(0.3)     # simulate slow write
    df.write_parquet(fp)

with tempfile.TemporaryDirectory() as tmp:
    tmp = Path(tmp)
    in_fps = [tmp / f"in_{i}.parquet" for i in range(2)]
    out_fp = tmp / "out.parquet"
    df0 = pl.DataFrame({"a": [1, 2]})
    df1 = pl.DataFrame({"a": [5, 6]})
    t = threading.Thread(target=slow_write, args=(df1, in_fps[1]))
    t.start()
    df0.write_parquet(in_fps[0])
    time.sleep(0.05)
    reduce_over(
        in_fps=in_fps, out_fp=out_fp,
        read_fn=pl.read_parquet,
        write_fn=pl.DataFrame.write_parquet,
        reduce_fn=lambda *dfs: pl.concat(dfs, how="vertical"),
        polling_time=0.01,
    )
    t.join()
```

Outputs:
```
polars.exceptions.ComputeError: parquet: File out of specification:
A parquet file must contain a header and footer with at least 12 bytes
```

## Prior Attempts

I reviewed the issue/PR history for related fixes. None of them address *this* race:

| # | Title | Merged | Addresses this bug? |
|---|-------|--------|---------------------|
| [#255](https://github.com/mmcdermott/MEDS_transforms/pull/255) | Use the `filelock` library for file locking | 2025-02-20 | No — hardens writer-vs-writer serialization, not reducer-vs-writer |
| [#346](https://github.com/mmcdermott/MEDS_transforms/issues/346) / [#347](https://github.com/mmcdermott/MEDS_transforms/pull/347) | Race condition in `rwlock_wrap`: lock file `unlink` FileNotFoundError | 2026-04-05 | No — fixes `lock_fp.unlink()` cleanup, a different race inside `rwlock_wrap.finally` |
| [#361](https://github.com/mmcdermott/MEDS_transforms/pull/361) | Remove redundant parquet writes in `write_for_test` | 2026-04-09 | No — test fixture cleanup, unrelated |
| [#144](https://github.com/mmcdermott/MEDS_transforms/issues/144) | Integration tests should run in parallel mode too | Closed 2024-08-29 | No — observed that "some race conditions only show up in parallel"; closed after adding one parallel test, but this particular race remained hidden because it's probabilistic |

**Why the previous fixes didn't catch this:** all prior work focused on the *writer side* (the `rwlock_wrap` / `FileLock` path and its cleanup). The reducer's polling loop was never treated as a locking participant — it just waits for file existence. With fast CI machines and tiny test files, the mapper usually finishes within one polling interval, so the race is rarely observed. When it is, it presents exactly as a `polars.ComputeError` about `PAR1`.

## Fix Options (for discussion, not yet implemented)

1. **Atomic publish on the writer**: in `rwlock_wrap`, have `write_fn` write to `out_fp.tmp` then `os.replace(out_fp.tmp, out_fp)`. Since POSIX `rename` is atomic, readers never see a partial file. Classically correct; subsumes (2) and (3).
2. **Reader-side completeness check**: replace `fp.is_file()` in `reduce_over` with the existing `is_complete_parquet_file` / `default_file_checker`.
3. **Reader waits for lockfile absence**: the reducer waits until `<fp>.lock` no longer exists — this is already a correct synchronization point because `rwlock_wrap.finally` releases the lock only after `write_fn` returns.

(1) is my preference — it's a single small change in `write_df` / `rwlock_wrap` and fixes the same class of bug for any future reader that doesn't know about lockfiles.

## Next Steps

- [ ] Add a deterministic regression test that captures this bug (initially as an `xfail`).
- [ ] Decide on fix strategy (probably (1) atomic rename).
- [ ] Re-enable the test once fixed.

I'll follow up with comments on the deep-dive / deterministic test.

mmcdermott:
## Summary

Fixes the mapreduce reducer race condition reported in #373. Originally this PR only added a failing regression test; PR #375 (now merged into this branch) added the fix, and a follow-up commit brought the patch to 100% line coverage.

### The bug

`reduce_over` used to poll for mapper outputs with `fp.is_file()`, which returns True as soon as the mapper *creates* `fp` — well before `df.write_parquet(fp)` has flushed a valid `PAR1` footer. The reducer then called `read_fn(fp)` on a partial parquet and raised:

```
polars.exceptions.ComputeError: parquet: File out of specification: The file must end with PAR1
```

### The fix — two complementary defenses

1. **Atomic writer-side publish** (`write_df`): stage to `<out_fp>.tmp`, then `os.replace` to `out_fp`. On POSIX (and same-filesystem Windows) the rename is atomic, so concurrent readers never observe a partial parquet at the final path. `except BaseException` unlinks the staging file so a mid-write failure leaves nothing behind.
2. **Completeness-aware reader-side poll** (`reduce_over`): the readiness loop now uses `default_file_checker`, which opens the parquet to verify the footer, instead of `Path.is_file()`. Defensive even if a caller-supplied `write_fn` isn't atomic.

The poll loop also grew an optional `max_poll_time` so a permanently-stuck / corrupt input produces a `TimeoutError` instead of hanging the reducer. `TimeoutError` messages distinguish `present but unreadable:` from `missing:` inputs.

### What's in this PR

| File | Change |
|---|---|
| `src/MEDS_transforms/dataframe/write_fn.py` | `write_df` now writes to `<out_fp>.tmp` and `os.replace`s to `out_fp`; cleans up staging file on any exception. |
| `src/MEDS_transforms/mapreduce/reducer.py` | `reduce_over` polls with `default_file_checker`, caches once-validated inputs, optional `max_poll_time` with bounded sleep (`min(polling_time, remaining)`), targeted `TimeoutError` messages. |
| `tests/test_reducer_race.py` | Four deterministic tests: the original race reproduction (now passing with the fix), timeout on a permanently-invalid parquet, timeout on a missing-entirely input, and the `max_poll_time <= polling_time` guard. |
| `tests/test_write_df_atomic.py` | Five tests: happy path, parent-dir creation, `LazyFrame` input, cleanup on `os.replace` failure, cleanup on mid-`write_parquet` failure. |

### Test plan

- [x] `uv run pytest tests/test_reducer_race.py tests/test_write_df_atomic.py`: **9 passed**
- [x] Full non-parallel suite (`tests/ + src/MEDS_transforms/`): **152 passed**
- [x] Coverage on the patch files: `write_fn.py` **100%**, `reducer.py` **100%**

Closes #373

### DIFF
--- a/src/MEDS_transforms/dataframe/write_fn.py
+++ b/src/MEDS_transforms/dataframe/write_fn.py
@@ -1,3 +1,4 @@
+import os
 from collections.abc import Callable
 from pathlib import Path
 
@@ -9,8 +10,20 @@ WRITE_FN_T = Callable[[DF_T, Path], None]
 
 
 def write_df(df: DF_T, out_fp: Path) -> None:
-    """A generic helper to write a dataframe, either lazy or eager, to a parquet file."""
+    """Atomically write a dataframe, either lazy or eager, to a parquet file.
+
+    Content is staged at ``<out_fp>.tmp`` and then ``os.replace``\\d onto
+    ``out_fp`` so concurrent readers never observe a partial parquet file at
+    the final path. The rename is atomic on POSIX and Windows as long as
+    ``out_fp`` and its staging path share a filesystem.
+    """
     if isinstance(df, pl.LazyFrame):
         df = df.collect()
     out_fp.parent.mkdir(parents=True, exist_ok=True)
-    df.write_parquet(out_fp, use_pyarrow=True)
+    tmp_fp = out_fp.with_suffix(out_fp.suffix + ".tmp")
+    try:
+        df.write_parquet(tmp_fp, use_pyarrow=True)
+        os.replace(tmp_fp, out_fp)
+    except BaseException:
+        tmp_fp.unlink(missing_ok=True)
+        raise
--- a/src/MEDS_transforms/dataframe/write_fn.py
+++ b/src/MEDS_transforms/dataframe/write_fn.py
@@ -12,8 +12,8 @@ WRITE_FN_T = Callable[[DF_T, Path], None]
 def write_df(df: DF_T, out_fp: Path) -> None:
     """Atomically write a dataframe, either lazy or eager, to a parquet file.
 
-    Content is staged at ``<out_fp>.tmp`` and then ``os.replace``\\d onto
-    ``out_fp`` so concurrent readers never observe a partial parquet file at
+    Content is staged at ``<out_fp>.tmp`` and then moved into place with
+    ``os.replace`` so concurrent readers never observe a partial parquet file at
     the final path. The rename is atomic on POSIX and Windows as long as
     ``out_fp`` and its staging path share a filesystem.
     """
--- a/src/MEDS_transforms/mapreduce/reducer.py
+++ b/src/MEDS_transforms/mapreduce/reducer.py
@@ -217,33 +217,39 @@ def reduce_over(
             "otherwise legitimately slow mappers will trigger a TimeoutError on the first recheck."
         )
 
-    deadline = time.monotonic() + max_poll_time if max_poll_time is not None else None
-    ready_fps: set[Path] = set()
+    deadline = None if max_poll_time is None else time.monotonic() + max_poll_time
+    ready: set[Path] = set()
+
+    def _is_ready(fp: Path) -> bool:
+        # ``is_file`` is a cheap stat; only if it passes do we open the parquet to check completeness.
+        return fp.is_file() and default_file_checker(fp)
+
     while True:
-        not_ready: list[Path] = []
-        for fp in in_fps:
-            if fp in ready_fps:
-                continue
-            # Fast path: skip the parquet completeness check when the file isn't even on disk yet.
-            if not fp.is_file() or not default_file_checker(fp):
-                not_ready.append(fp)
-                continue
-            ready_fps.add(fp)
-        if not not_ready:
+        ready.update(fp for fp in in_fps if fp not in ready and _is_ready(fp))
+        if len(ready) == len(in_fps):
             break
-        if deadline is not None and time.monotonic() >= deadline:
-            stuck = [fp for fp in not_ready if fp.exists()]
-            missing = [fp for fp in not_ready if not fp.exists()]
-            parts = []
-            if stuck:
-                parts.append(f"present but unreadable: {', '.join(str(fp) for fp in stuck)}")
-            if missing:
-                parts.append(f"missing: {', '.join(str(fp) for fp in missing)}")
-            raise TimeoutError(
-                f"Timed out after {max_poll_time}s waiting for reduction inputs — " + "; ".join(parts)
-            )
+
+        if deadline is None:
+            sleep_for = polling_time
+        else:
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                pending = [fp for fp in in_fps if fp not in ready]
+                stuck = [fp for fp in pending if fp.exists()]
+                missing = [fp for fp in pending if not fp.exists()]
+                parts = []
+                if stuck:
+                    parts.append(f"present but unreadable: {', '.join(str(fp) for fp in stuck)}")
+                if missing:
+                    parts.append(f"missing: {', '.join(str(fp) for fp in missing)}")
+                raise TimeoutError(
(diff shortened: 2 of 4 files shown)


mmcdermott:
Addressed the Copilot comments in abcb081:

- **On LazyFrame / `.equals()`** (valid): collected the `LazyFrame` before comparing and switched to `polars.testing.assert_frame_equal` for a clearer diff. The test still fails with the expected `ComputeError` on current `dev` because the race hits before the assertion — that's the intended signal of this PR.

- **On `xfail` vs unconditional fail** (push back): intentionally unconditional. The earlier version of this PR was `@pytest.mark.xfail(strict=True)`; it was changed on review precisely to surface the bug in CI so the fix has a concrete target (see issue #373 for the history). Marking it xfail again defeats that purpose. The PR body already calls out that CI will be red on merge until #375 lands, and the stacked-PR workflow resolves that within the same review cycle.

mmcdermott:
Round-2 addressed in 8857d4c:

- **`_reduce_fn` annotation** (valid): `read_df` returns a LazyFrame, so `reduce_over` passes LazyFrames into `_reduce_fn`. Updated to `pl.LazyFrame | pl.DataFrame`.

- **Flaky determinism from fixed sleep** (push back): the race window is 300ms. For the test to be flaky, the reducer would need to be delayed by >300ms after the `touched` event lands — on a machine where that happens, the whole suite is already in trouble. Making it event-deterministic would require a third signal from inside the reducer's `read_fn`, which complicates the test's story (the fix-path default_file_checker intentionally *never* calls `read_fn` on the partial file) without buying much. Leaving as-is; will revisit if we actually see flakes.

- **xfail suggestion** (push back — same as round 1): this test is intentionally unconditional on this branch. It pairs with PR #375 as a stacked fix; the two land together. The earlier version of this PR was `@pytest.mark.xfail(strict=True)`; it was changed on review to make the bug visible in CI (issue #373). Marking it xfail again defeats that purpose — the stacked #375 flips it green, and merging both at once keeps `dev` green.

mmcdermott:
Patch coverage pushed in b76e36e — both `write_fn.py` and `reducer.py` now at 100% locally.

- `reducer.py:244` (missing-input branch of the timeout error) → new `test_reduce_over_times_out_on_missing_input` covers it; the existing "permanently invalid" test also asserts the "present but unreadable" branch label.
- `write_fn.py:27-29` (tmp-file cleanup on failure) → new `tests/test_write_df_atomic.py` with five tests. Two of them simulate failures (one at `os.replace`, one mid-`write_parquet`) and assert both that the staging `.tmp` is cleaned up and that `out_fp` is left untouched.

PR title and description also updated to reflect that this branch now carries both the test and the fix (merged from #375). Dropping back to "Fix mapreduce reducer race condition (closes #373)".

Issue #56, opened 2024-07-24
We should support a "meta-stage" set-up where you can run multiple transformations in a single stage at once, without caching the files in between.


mmcdermott:
This would make things likely a fair bit faster and less disk expensive, especially when compiling multiple small transformations one after the other.

This can also tie into #55 in that both present fundamental improvements to the pipeline runnability and API.

mmcdermott:
## Proposed Action Plan: Meta-Stage Support

### Goal
Allow multiple transformations to run in a single stage without writing intermediate files to disk, improving both performance and disk usage.

### Approach

1. **Add a `composed_stage` or `meta_stage` concept to the Stage class** (`stages/base.py`):
   - Accept a list of stage names or stage functions in the pipeline config
   - Chain their `map_fn` callables together: output of one feeds directly into the next
   - Only write the final result to disk

2. **Pipeline config syntax** — something like:
   ```yaml
   stages:
     - compose:
         - add_time_derived_measurements
         - filter_measurements
         - reorder_measurements
   ```

3. **Implementation steps**:
   - Add a `ComposedStage` wrapper in `stages/base.py` that chains multiple `map_fn` callables
   - Update `PipelineConfig` in `configs/pipeline.py` to parse the `compose:` syntax
   - Update the runner to handle composed stages (single subprocess call, no intermediate dirs)
   - Ensure metadata stages can participate (may need to resolve #155 first, or restrict composition to data-only stages initially)

4. **Key constraint**: Only stages with compatible map-only semantics can be composed. Stages that require a reduce step (like `aggregate_code_metadata`) cannot participate.

### Complexity
Medium-high. The core chaining logic is simple, but integrating it cleanly with the pipeline config, runner, and stage registration system requires touching several files. Consider starting with a prototype that only supports data stages.

mmcdermott:
## Design follow-up for the flag-driven, no-user-code-changes direction

Building on the action plan above. Re-stating the goal per recent discussion:

- **User-facing**: a single CLI arg / pipeline flag. No per-pipeline rewrites.
- **Library-facing**: minimal surgery. Externally-registered stages must keep working.
- **Primary use case**: small datasets, or high-confidence pipelines where resumability isn't needed.

### Where the boundary actually is today

"In-memory" blurs three orthogonal concerns:

1. **Process boundary**: `runner.py` launches each stage as a subprocess (`MEDS_transform-stage pipeline.yaml stage=X`). Collapsing this means single-process execution.
[...]

PR #375, opened 2026-04-17
Fix reducer race: atomic write_df + completeness-aware poll

### LINKED ISSUE #373: Flaky CI: reducer reads partially-written parquet from mapper (PAR1 ComputeError)

mmcdermott:
## Summary

Fixes #373 with two complementary defenses against the mapreduce partial-parquet race:

1. **Writer side**: `write_df` now writes to `<out_fp>.tmp` and `os.replace`s onto `out_fp`. On POSIX (and Windows, same-filesystem), the rename is atomic — concurrent readers never observe a partial parquet at the final path. Covers mappers routed through `rwlock_wrap` and the reducer's own final write.
2. **Reader side**: `reduce_over`'s input-readiness poll now uses the existing `default_file_checker` (which opens the parquet to verify completeness) instead of `Path.is_file`. This keeps the reducer defensive even if a caller-supplied `write_fn` bypasses `write_df`.

Targets `fix/reducer-race-regression-test` so the failing test added in #374 turns green. Once merged, #374 can be reviewed and merged into `dev` with the fix and the regression test together.

## Why both fixes, not just one

- Writer-side alone: protects production readers of `write_df` outputs, but doesn't defend `reduce_over` against user-supplied `write_fn`s that aren't atomic. The failing test in this branch exercises exactly that case (synthetic non-atomic publisher via `fp.touch()`), so it would still fail.
- Reader-side alone: makes `reduce_over` robust, but leaves any other future reader of `write_df` outputs (e.g., downstream tools, user scripts) free to observe partial state.

Defense in depth is cheap here: `default_file_checker` already exists, and the atomic-rename pattern is a few lines.

## Test plan

- [x] Regression test from #374 (`test_reduce_over_waits_for_complete_parquet`) now passes reliably (10/10 trials, ~0.4s).
- [x] Full non-parallel suite (`uv run pytest tests/ src/MEDS_transforms/`): 144 passed.
- [ ] CI green on this PR.

Fixes #373

### DIFF
--- a/src/MEDS_transforms/dataframe/write_fn.py
+++ b/src/MEDS_transforms/dataframe/write_fn.py
@@ -1,3 +1,4 @@
+import os
 from collections.abc import Callable
 from pathlib import Path
 
@@ -9,8 +10,20 @@ WRITE_FN_T = Callable[[DF_T, Path], None]
 
 
 def write_df(df: DF_T, out_fp: Path) -> None:
-    """A generic helper to write a dataframe, either lazy or eager, to a parquet file."""
+    """Atomically write a dataframe, either lazy or eager, to a parquet file.
+
+    Content is staged at ``<out_fp>.tmp`` and then ``os.replace``\\d onto
+    ``out_fp`` so concurrent readers never observe a partial parquet file at
+    the final path. The rename is atomic on POSIX and Windows as long as
+    ``out_fp`` and its staging path share a filesystem.
+    """
     if isinstance(df, pl.LazyFrame):
         df = df.collect()
     out_fp.parent.mkdir(parents=True, exist_ok=True)
-    df.write_parquet(out_fp, use_pyarrow=True)
+    tmp_fp = out_fp.with_suffix(out_fp.suffix + ".tmp")
+    try:
+        df.write_parquet(tmp_fp, use_pyarrow=True)
+        os.replace(tmp_fp, out_fp)
+    except BaseException:
+        tmp_fp.unlink(missing_ok=True)
+        raise
--- a/src/MEDS_transforms/dataframe/write_fn.py
+++ b/src/MEDS_transforms/dataframe/write_fn.py
@@ -12,8 +12,8 @@ WRITE_FN_T = Callable[[DF_T, Path], None]
 def write_df(df: DF_T, out_fp: Path) -> None:
     """Atomically write a dataframe, either lazy or eager, to a parquet file.
 
-    Content is staged at ``<out_fp>.tmp`` and then ``os.replace``\\d onto
-    ``out_fp`` so concurrent readers never observe a partial parquet file at
+    Content is staged at ``<out_fp>.tmp`` and then moved into place with
+    ``os.replace`` so concurrent readers never observe a partial parquet file at
     the final path. The rename is atomic on POSIX and Windows as long as
     ``out_fp`` and its staging path share a filesystem.
     """
--- a/src/MEDS_transforms/mapreduce/reducer.py
+++ b/src/MEDS_transforms/mapreduce/reducer.py
@@ -217,33 +217,39 @@ def reduce_over(
             "otherwise legitimately slow mappers will trigger a TimeoutError on the first recheck."
         )
 
-    deadline = time.monotonic() + max_poll_time if max_poll_time is not None else None
-    ready_fps: set[Path] = set()
+    deadline = None if max_poll_time is None else time.monotonic() + max_poll_time
+    ready: set[Path] = set()
+
+    def _is_ready(fp: Path) -> bool:
+        # ``is_file`` is a cheap stat; only if it passes do we open the parquet to check completeness.
+        return fp.is_file() and default_file_checker(fp)
+
     while True:
-        not_ready: list[Path] = []
-        for fp in in_fps:
-            if fp in ready_fps:
-                continue
-            # Fast path: skip the parquet completeness check when the file isn't even on disk yet.
-            if not fp.is_file() or not default_file_checker(fp):
-                not_ready.append(fp)
-                continue
-            ready_fps.add(fp)
-        if not not_ready:
+        ready.update(fp for fp in in_fps if fp not in ready and _is_ready(fp))
+        if len(ready) == len(in_fps):
             break
-        if deadline is not None and time.monotonic() >= deadline:
-            stuck = [fp for fp in not_ready if fp.exists()]
-            missing = [fp for fp in not_ready if not fp.exists()]
-            parts = []
-            if stuck:
-                parts.append(f"present but unreadable: {', '.join(str(fp) for fp in stuck)}")
-            if missing:
-                parts.append(f"missing: {', '.join(str(fp) for fp in missing)}")
-            raise TimeoutError(
-                f"Timed out after {max_poll_time}s waiting for reduction inputs — " + "; ".join(parts)
-            )
+
+        if deadline is None:
+            sleep_for = polling_time
+        else:
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                pending = [fp for fp in in_fps if fp not in ready]
+                stuck = [fp for fp in pending if fp.exists()]
+                missing = [fp for fp in pending if not fp.exists()]
+                parts = []
+                if stuck:
+                    parts.append(f"present but unreadable: {', '.join(str(fp) for fp in stuck)}")
+                if missing:
+                    parts.append(f"missing: {', '.join(str(fp) for fp in missing)}")
+                raise TimeoutError(
+                    f"Timed out after {max_poll_time}s waiting for reduction inputs — " + "; ".join(parts)
+                )
+            # Don't oversleep past the deadline; if remaining < polling_time, wake up sooner.
+            sleep_for = min(polling_time, remaining)
+
         logger.info("Waiting to begin reduction for all files to be written...")
-        time.sleep(polling_time)
+        time.sleep(sleep_for)
 
     reduced = reduce_fn(*[read_fn(fp) for fp in in_fps])
 
--- a/src/MEDS_transforms/mapreduce/reducer.py
+++ b/src/MEDS_transforms/mapreduce/reducer.py
@@ -29,7 +29,7 @@ def reduce_over(
     merge_fn: REDUCE_FN_T | None = None,
     do_overwrite: bool = False,
     polling_time: float = 0.1,
-    max_poll_time: float = 300.0,
+    max_poll_time: float | None = None,
 ):
     """Performs a reduction operation on a list of input file paths, with optional merging to existing data.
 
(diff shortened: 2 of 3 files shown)


mmcdermott:
Addressed both Copilot comments in a171979:
- docstring no longer has `\d`; rephrased as 'moved into place with os.replace'.
- `reduce_over` accepts `max_poll_time` (default 300s) and raises `TimeoutError` when inputs stay unreadable past the deadline, distinguishing present-but-corrupt from missing. New regression test `test_reduce_over_times_out_on_permanently_invalid_input` covers it.

mmcdermott:
Addressed the second-round Copilot comments in 90cde1d:

**On default `max_poll_time=300` conflicting with `cfg.polling_time=300`**: valid catch — matching defaults would have caused spurious timeouts on the first recheck of any legitimately slow mapper. Flipped the default to `None` (no timeout) so timeouts are opt-in, and added a `ValueError` when a caller sets `max_poll_time <= polling_time`.

**On the poll calling `default_file_checker` every iteration for every input**: valid — for N shards we were opening N parquet files each iteration. Added a `ready_fps` set that caches once-validated inputs, plus a cheap `Path.is_file()` short-circuit before the parquet open. Pending files now cost one stat each; already-ready files cost nothing.

New test `test_reduce_over_rejects_max_poll_time_not_larger_than_polling_time` covers the new guard.

src/MEDS_transforms/dataframe/write_fn.py
```diff
@@ -9,8 +10,20 @@
...
+    ``out_fp`` and its staging path share a filesystem.
+    """
     if isinstance(df, pl.LazyFrame):
         df = df.collect()
     out_fp.parent.mkdir(parents=True, exist_ok=True)
-    df.write_parquet(out_fp, use_pyarrow=True)
+    tmp_fp = out_fp.with_suffix(out_fp.suffix + ".tmp")
+    try:
+        df.write_parquet(tmp_fp, use_pyarrow=True)
+        os.replace(tmp_fp, out_fp)
```
mmcdermott, line 26: Does this have any performance implications? Is there not a library that exists for python that can be used to do this that may be faster? Please push back if I misunderstand.

src/MEDS_transforms/mapreduce/reducer.py
```diff
@@ -203,7 +211,37 @@ def reduce_over(
...
-    while not all(fp.is_file() for fp in in_fps):
+    if max_poll_time is not None and max_poll_time <= polling_time:
+        raise ValueError(
+            f"max_poll_time ({max_poll_time}s) must be greater than polling_time ({polling_time}s); "
+            "otherwise legitimately slow mappers will trigger a TimeoutError on the first recheck."
+        )
+
+    deadline = time.monotonic() + max_poll_time if max_poll_time is not None else None
+    ready_fps: set[Path] = set()
+    while True:
```
mmcdermott, line 227: This loop seems needlessly convoluted with some indirection. Is there not a simpler way to do this, using some python parallelization primitives? Or, if that exposes more risks, can this code just be cleaned up, to avoid so many "not X" indirects and continues and such?

tests/test_reducer_race.py
```diff
@@ -83,6 +83,51 @@ def test_reduce_over_waits_for_complete_parquet() -> None:
...
+                write_fn=write_df,
+                reduce_fn=_reduce_fn,
+                polling_time=0.01,
+                max_poll_time=0.3,
+            )
+
+
+def test_reduce_over_rejects_max_poll_time_not_larger_than_polling_time() -> None:
+    """``max_poll_time`` must exceed ``polling_time`` to avoid spurious timeouts on the first poll."""
+    import pytest
```
mmcdermott: Do this at the module level, if it is even necessary.

tests/test_reducer_race.py
```diff
@@ -83,6 +83,51 @@ def test_reduce_over_waits_for_complete_parquet() -> None:
...
+        result = read_df(out_fp).collect().sort("a")
         expected = pl.concat([df0, df1], how="vertical").sort("a")
         assert result.equals(expected), f"Reducer output differs:\n{result}\nvs expected:\n{expected}"
+
+
+def test_reduce_over_times_out_on_permanently_invalid_input() -> None:
+    """A permanently invalid (empty) parquet should time out, not hang forever."""
+    import pytest
+
+    with tempfile.TemporaryDirectory() as tmp:
```
mmcdermott: could a tmppath fixture be useful here? Or the yaml_to_disk utility? Here and below.

src/MEDS_transforms/dataframe/write_fn.py
mmcdermott: Pushing back — this stdlib pattern is the canonical one and the perf cost is trivial.

- `os.replace` on the same filesystem is a single rename syscall — microseconds, fixed cost regardless of file size. The per-file overhead is negligible relative to the parquet write itself.
- The external libraries in this space (`python-atomicwrites`, `python-atomicwrites-homeassistant`) implement *exactly this pattern* — write to a sibling tempfile, then `os.replace`. They're ~50 lines of wrapper around the same two stdlib calls. Adding a dependency would be pure overhead.
- `pyarrow` and `polars` don't ship an atomic-write option at all (as documented in the research I posted on #56 / the design discussion); both projects tell you to stage externally.

If anything, I'd argue this is the lightest possible implementation — happy to reconsider if you have a specific library in mind.

src/MEDS_transforms/mapreduce/reducer.py
mmcdermott: Good call — rewritten in b8d1c9b. The new loop uses a positive `ready` set fed by an `_is_ready` helper, an explicit `len(ready) == len(in_fps)` loop condition, and a single branch for deadline/sleep accounting. No more `not_ready` list, no more `not fp.is_file() or not …` double-negative.

On the parallelization-primitives question: I don't think it buys us anything here. The readiness signal for a file comes from "the filesystem has a valid parquet now," which has no kernel-level notify we can wait on (inotify doesn't fire on remote filesystems / NFS, and doesn't distinguish empty-touch from complete-write anyway). Polling is genuinely the right shape; the cleanup is about how the poll is expressed.

tests/test_reducer_race.py
mmcdermott: Fixed in b8d1c9b — `import pytest` pulled to module level.

tests/test_reducer_race.py
mmcdermott: Good call — switched to pytest's `tmp_path` fixture across all three tests in b8d1c9b. `yaml_to_disk` wasn't a natural fit here since we're writing parquet (not YAML scaffolding), but `tmp_path` removes the boilerplate cleanly.

mmcdermott:
Addressed all 4 user comments plus the round-2 Copilot comment in b8d1c9b (branch force-pushed after rebasing on top of the refreshed #374 test). Per-comment replies posted inline above.

Summary:
- **Perf/library question on `write_fn.py`**: pushed back — os.replace is the canonical atomic-publish pattern and external libs are thin wrappers over the same two syscalls.
- **Convoluted poll loop**: cleaned up; positive `ready` set, helper function, single deadline branch, no more `not_ready`/double-negatives.
- **`import pytest` in function body**: moved to module level.
- **`tmp_path` fixture**: adopted across all three tests.
- **Copilot round-2 on sleep overshooting the deadline**: fixed with `min(polling_time, remaining)`.

Merge conflict with `fix/reducer-race-regression-test` resolved via rebase — branch now applies cleanly.

tests/test_reducer_race.py
mmcdermott: yaml_to_disk (https://github.com/mmcdermott/yaml_to_disk) let's you express what you want to write to a temp path on disk in yaml form, to write it concisely and clearly. It doesn't do what you thought it did. But, I don't think you need it here.

PR #376, opened 2026-04-17
Strip pipeline-only overrides before forwarding to stages (closes #371)

### LINKED ISSUE #371: Pipeline-level `--overrides` are forwarded verbatim to stage invocations, breaking `parallelize.*` (and any other pipeline-only key) overrides
## Summary

In MEDS-Transforms 0.6.6, passing a pipeline-level key like \`parallelize.*\` to \`MEDS_transform-pipeline --overrides ...\` does not work. The outer pipeline config merges correctly, but the same CLI overrides are then forwarded into every inner \`MEDS_transform-stage\` invocation, where the stage's strict Hydra struct rejects keys it doesn't know about. This means downstream wrappers can't offer override-based parallelism UX — the only working paths are \`--stage_runner_fp\` or baking \`parallelize:\` into the pipeline YAML file.

## Reproduction

Using MEDS-Transforms 0.6.6 against any pipeline config that does **not** already include a \`parallelize:\` block (e.g., \`meds-torch-data\`'s \`_MTD_preprocess.yaml\`):

### (1) Without a prefix — rejected by inner stage

\`\`\`bash
INPUT_DIR=/path/in OUTPUT_DIR=/path/out \\
  MEDS_transform-pipeline /path/to/pipeline.yaml \\
  --overrides parallelize.n_workers=2 parallelize.launcher=joblib
\`\`\`

Outer pipeline log:
\`\`\`
INFO:MEDS_transforms.runner:Parallelization configuration loaded from pipeline config
INFO:MEDS_transforms.runner:Running command: MEDS_transform-stage ... --multirun stage=fit_normalization parallelize.n_workers=2 parallelize.launcher=joblib worker=\"range(0,2)\" hydra/launcher=joblib
\`\`\`

Inner stage fails:
\`\`\`
Could not override 'parallelize.n_workers'.
To append to your config use +parallelize.n_workers=2
Key 'parallelize' is not in struct
    full_key: parallelize
    object_type=dict
\`\`\`

### (2) With \`+\` prefix — silently runs serial

\`\`\`bash
INPUT_DIR=/path/in OUTPUT_DIR=/path/out \\
  MEDS_transform-pipeline /path/to/pipeline.yaml \\
  --overrides +parallelize.n_workers=2 +parallelize.launcher=joblib
\`\`\`

Returncode 0 — but log shows:
\`\`\`
INFO:MEDS_transforms.runner:No parallelization configuration provided.
INFO:MEDS_transforms.runner:Running command: MEDS_transform-stage ... fit_normalization stage=fit_normalization +parallelize.n_workers=2 +parallelize.launcher=joblib
\`\`\`

No multirun, no joblib, no parallelism — the pipeline ran sequentially. The reason: \`OmegaConf.from_dotlist\` ([configs/pipeline.py#L319-L320](https://github.com/mmcdermott/MEDS_transforms/blob/main/src/MEDS_transforms/configs/pipeline.py)) treats \`+\` as a literal character, so the merged config carries a key named \`\"+parallelize\"\` (not \`parallelize\`). The runner's \`\"parallelize\" in additional_params\` check ([runner.py#L344](https://github.com/mmcdermott/MEDS_transforms/blob/main/src/MEDS_transforms/runner.py)) misses it entirely. Inner stages then accept \`+parallelize.*\` as Hydra append-keys and they're never consumed. Worst-of-both-worlds: looks successful but silently ignored.

## Root cause

\`runner.py\`'s \`main\` calls \`run_stage(..., cfg_overrides=args.overrides, ...)\`. Then \`run_stage\` does:

\`\`\`python
if cfg_overrides:
    command_parts.extend(cfg_overrides)
\`\`\`

This forwards **every** CLI override from the outer pipeline call into the inner \`MEDS_transform-stage\` command, regardless of whether the key is a pipeline-level concern (\`parallelize.*\`, \`stages\`, etc.) or a stage-level concern (\`do_overwrite\`, \`stage.*\`, etc.). Pipeline-level keys don't exist in the stage's structured config, so Hydra rejects them.

## Desired behavior

At least one of:

1. **Filter pipeline-level overrides before forwarding to stages.** Known pipeline-level keys (\`parallelize.*\`, \`stages\`, etc.) should be consumed by the outer runner and not propagated to \`MEDS_transform-stage\`.
2. **Make inner stage configs non-struct at the root** so unknown keys are tolerated (this is a bigger behavior change and may mask real user typos).
3. **Document the limitation and direct users to \`stage_runner_fp\`** as the only supported path for pipeline-level configuration overrides.

Option (1) is the cleanest — it matches the mental model that \`--overrides\` applies to the pipeline config, and the pipeline runner is already the authority on what's pipeline-level vs. stage-level.

## Impact on downstream

[meds-torch-data](https://github.com/mmcdermott/meds-torch-data) wraps \`MEDS_transform-pipeline\` for MEDS → PyTorch preprocessing. We've been trying to offer a first-class parallelism knob on our entrypoint that compiles to inner \`parallelize.*\` overrides (see [mmcdermott/meds-torch-data#84](https://github.com/mmcdermott/meds-torch-data/issues/84) for the full design discussion). That plan is blocked by this bug. Our current workaround: require users to hand-author a \`stage_runner.yaml\` and pass it via \`--stage_runner_fp\`. It works but duplicates the \`parallelize:\` contract across two places (stage runner vs. pipeline YAML) and prevents a single-knob CLI UX.

## Version

\`MEDS_transforms==0.6.6\`, \`hydra-core>=1.3\`, \`omegaconf>=2.3\`, Python 3.12. Reproduced on Linux.

mmcdermott:
## Summary

Fixes #371. ``MEDS_transform-pipeline --overrides`` previously forwarded every override verbatim to each inner ``MEDS_transform-stage`` invocation. Pipeline-only keys like ``parallelize.*`` are not in the stage's structured Hydra config, so the stage rejected them:

```
Could not override 'parallelize.n_workers'. Key 'parallelize' is not in struct
```

Now the runner filters pipeline-only roots (``parallelize``, ``stages``, ``description``) out of the forwarded overrides so:

```bash
MEDS_transform-pipeline pipeline.yaml --overrides parallelize.n_workers=2 parallelize.launcher=joblib
```

actually works end-to-end.

## Test plan

- [x] New doctests for ``_override_root_key`` and ``_stage_forwardable_overrides`` cover ``+``, ``++``, ``~`` prefix operators and mixed pipeline/stage overrides.
- [x] Extended ``run_stage`` doctest shows ``parallelize.*`` stripped and ``do_overwrite=True`` preserved in the generated command.
- [x] Full non-parallel suite: 145 passed.

Closes #371

### DIFF
--- a/src/MEDS_transforms/runner.py
+++ b/src/MEDS_transforms/runner.py
@@ -25,6 +25,60 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+# Override keys whose root is consumed by the pipeline runner and must not be forwarded to inner
+# ``MEDS_transform-stage`` invocations. Forwarding pipeline-only keys makes Hydra reject the inner
+# command because the stage's structured config does not declare them. See issue #371.
+PIPELINE_ONLY_OVERRIDE_ROOTS: frozenset[str] = frozenset({"parallelize", "stages", "description"})
+
+
+def _override_root_key(override: str) -> str:
+    """Extract the root config key targeted by a Hydra-style CLI override string.
+
+    Strips any leading Hydra prefix operators (``+``, ``++``, ``~``) and returns the segment before
+    the first ``.`` or ``=``.
+
+    Examples:
+        >>> _override_root_key("parallelize.n_workers=2")
+        'parallelize'
+        >>> _override_root_key("+parallelize.launcher=joblib")
+        'parallelize'
+        >>> _override_root_key("++parallelize.launcher_params.foo=bar")
+        'parallelize'
+        >>> _override_root_key("~parallelize.n_workers")
+        'parallelize'
+        >>> _override_root_key("do_overwrite=True")
+        'do_overwrite'
+        >>> _override_root_key("stage_cfg.min_events_per_subject=5")
+        'stage_cfg'
+    """
+    stripped = override.lstrip("+~")
+    if "=" in stripped:
+        stripped = stripped.split("=", 1)[0]
+    return stripped.split(".", 1)[0]
+
+
+def _stage_forwardable_overrides(cfg_overrides: list[str] | None) -> list[str]:
+    """Return the subset of ``cfg_overrides`` that is safe to forward to a stage subprocess.
+
+    Pipeline-only keys (``parallelize``, ``stages``, ``description``) are consumed by the outer
+    runner and would be rejected by the stage's strict Hydra struct. See issue #371.
+
+    Examples:
+        >>> _stage_forwardable_overrides(None)
+        []
+        >>> _stage_forwardable_overrides([])
+        []
+        >>> _stage_forwardable_overrides(["parallelize.n_workers=2", "do_overwrite=True"])
+        ['do_overwrite=True']
+        >>> _stage_forwardable_overrides(["+parallelize.launcher=joblib", "stage_cfg.x=1"])
+        ['stage_cfg.x=1']
+        >>> _stage_forwardable_overrides(["stages=[a,b]", "description=hello", "ok=true"])
+        ['ok=true']
+    """
+    if not cfg_overrides:
+        return []
+    return [o for o in cfg_overrides if _override_root_key(o) not in PIPELINE_ONLY_OVERRIDE_ROOTS]
+
 
 def get_parallelization_args(
     parallelization_cfg: dict | DictConfig | None, default_parallelization_cfg: dict | DictConfig
@@ -207,6 +261,25 @@ def run_stage(
             ...
         ValueError: Stage reorder_measurements failed via ...
 
+        Pipeline-only overrides (``parallelize.*``, ``stages``, ``description``) are stripped before
+        the stage command is assembled so the stage's strict Hydra struct does not reject them:
+
+        >>> stage_runners["reorder_measurements"].pop("parallelize", None)
+        {'n_workers': 2}
+        >>> run_stage(
+        ...     "pipeline_config.yaml",
+        ...     stage_runners,
+        ...     pipeline_cfg,
+        ...     "reorder_measurements",
+        ...     cfg_overrides=[
+        ...         "parallelize.n_workers=4",
+        ...         "+parallelize.launcher=joblib",
+        ...         "do_overwrite=True",
+        ...     ],
+        ...     runner_fn=fake_shell_succeed,
+        ... )
+        baz_script stage=reorder_measurements do_overwrite=True
+
         Improper configurations also raise errors:
 
         >>> bad_runners = {"reshard_to_split": {"_base_stage": "belongs in the stage"}}
@@ -244,8 +317,9 @@ def run_stage(
         f"stage={stage_name}",
     ]
 
-    if cfg_overrides:
-        command_parts.extend(cfg_overrides)
+    stage_overrides = _stage_forwardable_overrides(cfg_overrides)
+    if stage_overrides:
+        command_parts.extend(stage_overrides)
 
     parallelization_args = get_parallelization_args(
         stage_runner_config.get("parallelize", {}), default_parallelization_cfg


mmcdermott:
Do we really need to fix this? This issue doesn't seem like it is actually a major blocker of anything, as stage yaml files can be used in the interim, and we don't have a test that is failing due to this behavior. In this PR, it also looks like we are discarding those parameters, and then not using them for anything, which may be even worse than throwing an error, right? This seems like maybe not the right solution.

mmcdermott:
Evaluating your pushback — I think you're substantially right. Here's my read:

## Is the bug real?

Yes — issue #371 documents a concrete downstream motivation (`meds-torch-data` wanting a single-knob parallelism CLI) and two distinct failure modes:

1. **Plain key** (`parallelize.n_workers=2`): Hydra rejects at the inner stage's strict struct.
2. **`+`-prefixed** (`+parallelize.n_workers=2`): returncode 0, but silently runs serial.

## What this PR actually fixes

Only case (1). Trace:

- `PipelineConfig.from_arg` calls `OmegaConf.from_dotlist(cfg_overrides)` (pipeline.py:320) — which does **not** understand Hydra's `+`/`++`/`~` prefixes; it treats them as literal characters.
- For plain `parallelize.n_workers=2`: `from_dotlist` produces `{parallelize: {n_workers: 2}}`, which merges into `additional_params["parallelize"]`, which runner.py:344 picks up as `default_parallelization_cfg`. So the outer override *does* take effect — and filtering pipeline-only roots out of the stage forward is the correct completion of that path.
- For `+parallelize.n_workers=2`: `from_dotlist` produces a key literally named `"+parallelize"`. The `"parallelize" in additional_params` check misses it. With this PR, it's now *also* stripped from the stage forward. **Net effect: still silently ignored, just silent in a different place.**

So your concern — *"discarding those parameters, and then not using them for anything"* — is exactly right for the `+`/`++`/`~` prefix cases. This PR doesn't make that worse than today's behavior, but it doesn't fix it either. And there's no test that exercises end-to-end that `--overrides parallelize.n_workers=2` actually results in 2 workers.

## What's missing

1. `PipelineConfig.from_arg` should be Hydra-prefix-aware (strip `+`/`++`/`~` before `from_dotlist`, or use Hydra's own override parser), so *all* prefix variants merge into `additional_params` correctly at the outer level.
2. An end-to-end test: invoke the runner with `cfg_overrides=["parallelize.n_workers=3"]` against a pipeline YAML without a `parallelize:` block, and assert the emitted stage command includes `--multirun worker="range(0,3)"`. Without that, the PR is asserting a negative (keys removed) rather than a positive (override honored).
3. Arguably, any pipeline-only override the outer runner can't place (e.g. `stages=...` as a dotlist, or a prefix-keyed `parallelize.*` that didn't merge) should warn rather than vanish.

## Interim workaround

Your point about stage yaml files / baking `parallelize:` into the pipeline YAML stands — both paths work today. The whole issue is really about UX polish for wrappers like `meds-torch-data`, not a correctness blocker for anyone who can touch the YAMLs.

## Recommendation

Either:

- **(a) Close this PR and defer** until the outer runner parses Hydra prefixes properly — that's the load-bearing fix; this filter is just cleanup on top of it. Cleaner to do together.
- **(b) Expand this PR** to cover (1) prefix-aware merging in `from_arg`, (2) an end-to-end "override is honored" doctest, and (3) a warning when a pipeline-only override gets dropped without effect. Then the filter is the last mile of a complete fix rather than a fig-leaf over a half-working path.

I'd lean (a) — the surface area grows fast once you try to do this right, and the `meds-torch-data` consumer has a functional (if ugly) `stage_runner_fp` path today.

mmcdermott:
Closing unmerged. As laid out above, this PR only addresses the plain-key failure mode of #371 and leaves the `+`/`++`/`~` prefix cases silently discarded — which, as you noted, is arguably worse than the current Hydra-rejects behavior. The real fix lives one layer up in `PipelineConfig.from_arg` (making `--overrides` Hydra-prefix-aware), and without that the filter here is a fig leaf on a half-working path.

Deferring the whole thing — `stage_runner_fp` and inline `parallelize:` in the pipeline YAML remain the supported paths for now. Closing #371 as not-planned; we can re-open if/when the outer-runner Hydra-override story gets picked up.

mmcdermott:
Closing unmerged — see discussion above and on #371. Deferring pending a proper Hydra-prefix-aware fix in `PipelineConfig.from_arg`.

PR #377, opened 2026-04-17
Dynamic quantile schema for aggregate_code_metadata (closes #342)

### LINKED ISSUE #342: This isn't right; the output will depend upon the passed probs to the quantile call...
https://github.com/mmcdermott/MEDS_transforms/blob/9af69de437015e8192b6f2ffb5642a7982e2ded8/src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py#L790

mmcdermott:
## Summary

Fixes #342. The static ``AGGREGATION_SCHEMA_UPDATES`` dict hardcoded the quantile struct to ``0.25/0.5/0.75``, so users configuring non-default ``probs`` would produce an output whose expected-schema validation inside ``StageExample`` failed (the reducer itself was correct; only the schema override was wrong).

## Changes

1. ``Stage.output_schema_updates`` now accepts either a ``dict`` or a callable ``(stage_cfg) -> dict``. A new ``Stage._resolve_output_schema_updates(example_dir=None)`` helper centralises resolution.
2. ``aggregate_code_metadata`` builds the ``values/quantiles`` ``pl.Struct`` from the example's declared aggregations. Default quantiles continue to produce the same schema as before; non-default ``probs`` now validate correctly.
3. ``Stage.__str__`` and ``docgen.generate_stage_docs`` route through the shared resolver so doc output and pretty-print use the resolved schema.

## Test plan

- [x] New doctests on ``_quantiles_schema`` and ``aggregation_schema_updates`` cover both default and non-default probs.
- [x] Full non-parallel suite: **145 passed**.
- [x] All existing ``aggregate_code_metadata`` examples continue to validate with the dynamic schema.

Closes #342

### DIFF
--- a/src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
+++ b/src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
@@ -788,14 +788,7 @@ def reducer_fntr(
     return reducer
 
 
-AGGREGATION_SCHEMA_UPDATES = {
-    "values/quantiles": pl.Struct(
-        {
-            "values/quantile/0.25": pl.Float32,
-            "values/quantile/0.5": pl.Float32,
-            "values/quantile/0.75": pl.Float32,
-        }
-    ),
+AGGREGATION_SCHEMA_UPDATES_BASE = {
     "code/n_occurrences": pl.UInt8,
     "code/n_subjects": pl.UInt8,
     "values/n_occurrences": pl.UInt8,
@@ -808,6 +801,52 @@ AGGREGATION_SCHEMA_UPDATES = {
 }
 
 
+def _quantiles_schema(aggregations: list) -> pl.Struct | None:
+    """Build the ``values/quantiles`` struct schema for the quantile list declared in ``aggregations``.
+
+    Returns ``None`` when no ``values/quantiles`` aggregation is declared.
+
+    Examples:
+        >>> _quantiles_schema(["code/n_occurrences"]) is None
+        True
+        >>> s = _quantiles_schema([{"name": "values/quantiles", "quantiles": [0.1, 0.5, 0.9]}])
+        >>> list(s.fields)
+        [Field('values/quantile/0.1', Float32), Field('values/quantile/0.5', Float32), Field('values/quantile/0.9', Float32)]
+        >>> s = _quantiles_schema([{"name": "values/quantiles", "quantiles": [0.25, 0.75]}])
+        >>> list(s.fields)
+        [Field('values/quantile/0.25', Float32), Field('values/quantile/0.75', Float32)]
+    """  # noqa: E501
+    for agg in aggregations or []:
+        if isinstance(agg, dict) and agg.get("name") == "values/quantiles":
+            qs = agg.get("quantiles", []) or []
+            return pl.Struct({f"values/quantile/{q}": pl.Float32 for q in qs})
+    return None
+
+
+def aggregation_schema_updates(stage_cfg: dict | None = None) -> dict[str, pl.DataType]:
+    """Compose the example-time output schema overrides for this stage, including dynamic quantiles.
+
+    Static column types (``code/n_occurrences``, ``values/sum``, etc.) are always emitted. The
+    ``values/quantiles`` struct is generated from the example's ``aggregations`` config so that
+    non-default ``probs`` round-trip correctly (see issue #342).
+
+    Examples:
+        >>> aggregation_schema_updates({"aggregations": ["code/n_occurrences"]})["values/sum"]
+        Float32
+        >>> "values/quantiles" in aggregation_schema_updates({"aggregations": ["code/n_occurrences"]})
+        False
+        >>> cfg = {"aggregations": [{"name": "values/quantiles", "quantiles": [0.1, 0.9]}]}
+        >>> list(aggregation_schema_updates(cfg)["values/quantiles"].fields)
+        [Field('values/quantile/0.1', Float32), Field('values/quantile/0.9', Float32)]
+    """
+    updates = dict(AGGREGATION_SCHEMA_UPDATES_BASE)
+    if stage_cfg:
+        quantiles_struct = _quantiles_schema(stage_cfg.get("aggregations"))
+        if quantiles_struct is not None:
+            updates["values/quantiles"] = quantiles_struct
+    return updates
+
+
 stage = Stage.register(
-    map_fn=mapper_fntr, reduce_fn=reducer_fntr, output_schema_updates=AGGREGATION_SCHEMA_UPDATES
+    map_fn=mapper_fntr, reduce_fn=reducer_fntr, output_schema_updates=aggregation_schema_updates
 )
--- a/src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
+++ b/src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
@@ -833,7 +833,8 @@ def aggregation_schema_updates(stage_cfg: dict | None = None) -> dict[str, pl.Da
 
     Static column types (``code/n_occurrences``, ``values/sum``, etc.) are always emitted. The
     ``values/quantiles`` struct is generated from the example's ``aggregations`` config so that
-    non-default ``probs`` round-trip correctly (see issue #342).
+    non-default ``quantiles`` (i.e. the quantile probabilities) round-trip correctly (see
+    issue #342).
 
     Examples:
         >>> from pprint import pprint
--- a/src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
+++ b/src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
@@ -810,12 +810,17 @@ def _quantiles_schema(aggregations: list) -> pl.Struct | None:
         >>> _quantiles_schema(["code/n_occurrences"]) is None
         True
         >>> s = _quantiles_schema([{"name": "values/quantiles", "quantiles": [0.1, 0.5, 0.9]}])
-        >>> list(s.fields)
-        [Field('values/quantile/0.1', Float32), Field('values/quantile/0.5', Float32), Field('values/quantile/0.9', Float32)]
+        >>> for f in s.fields:
+        ...     print(f)
+        Field('values/quantile/0.1', Float32)
+        Field('values/quantile/0.5', Float32)
+        Field('values/quantile/0.9', Float32)
         >>> s = _quantiles_schema([{"name": "values/quantiles", "quantiles": [0.25, 0.75]}])
-        >>> list(s.fields)
-        [Field('values/quantile/0.25', Float32), Field('values/quantile/0.75', Float32)]
-    """  # noqa: E501
+        >>> for f in s.fields:
+        ...     print(f)
+        Field('values/quantile/0.25', Float32)
+        Field('values/quantile/0.75', Float32)
+    """
     for agg in aggregations or []:
         if isinstance(agg, dict) and agg.get("name") == "values/quantiles":
             qs = agg.get("quantiles", []) or []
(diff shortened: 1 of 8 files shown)


mmcdermott:
Both comments addressed in 2b4baa5:

- **default_config fallback**: _resolve_output_schema_updates now falls back to self.default_config when no explicit example_dir or pre-parsed stage_cfg is provided, so Stage.__str__ and docgen reflect the stage's default behavior.
- **Double-read of cfg.yaml**: Stage.test_cases now parses cfg.yaml once and shares the dict with both the schema resolver (new keyword-only stage_cfg= arg) and StageExample.from_dir (new keyword-only stage_cfg= arg). Both signatures are backwards-compatible — existing callers pass only example_dir and keep the old behavior.

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
```diff
@@ -808,6 +801,52 @@ def reducer(*dfs: Sequence[pl.LazyFrame]) -> pl.LazyFrame:
...
+    Examples:
+        >>> _quantiles_schema(["code/n_occurrences"]) is None
+        True
+        >>> s = _quantiles_schema([{"name": "values/quantiles", "quantiles": [0.1, 0.5, 0.9]}])
+        >>> list(s.fields)
+        [Field('values/quantile/0.1', Float32), Field('values/quantile/0.5', Float32), Field('values/quantile/0.9', Float32)]
+        >>> s = _quantiles_schema([{"name": "values/quantiles", "quantiles": [0.25, 0.75]}])
+        >>> list(s.fields)
+        [Field('values/quantile/0.25', Float32), Field('values/quantile/0.75', Float32)]
+    """  # noqa: E501
```
mmcdermott: Why noqa?

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
```diff
@@ -808,6 +801,52 @@ def reducer(*dfs: Sequence[pl.LazyFrame]) -> pl.LazyFrame:
...
+
+def aggregation_schema_updates(stage_cfg: dict | None = None) -> dict[str, pl.DataType]:
+    """Compose the example-time output schema overrides for this stage, including dynamic quantiles.
+
+    Static column types (``code/n_occurrences``, ``values/sum``, etc.) are always emitted. The
+    ``values/quantiles`` struct is generated from the example's ``aggregations`` config so that
+    non-default ``probs`` round-trip correctly (see issue #342).
+
+    Examples:
+        >>> aggregation_schema_updates({"aggregations": ["code/n_occurrences"]})["values/sum"]
```
mmcdermott: Can you just show the output, instead of accessing one specific key? It will make a better doctest.

src/MEDS_transforms/stages/base.py
```diff
@@ -721,6 +727,73 @@ def default_config(self, default_config: dict[str, Any] | DictConfig | Path | st
...
+            ...     default_config={"probs": "from-default"},
+            ... )
+            >>> stage._resolve_output_schema_updates(stage_cfg={"probs": "from-arg"})
+            {'probs': 'from-arg'}
+            >>> import tempfile
+            >>> with tempfile.TemporaryDirectory() as td:
+            ...     example_dir = Path(td)
+            ...     (example_dir / "cfg.yaml").write_text("probs: from-example\\n")
+            ...     stage._resolve_output_schema_updates(example_dir)
+            20
```
mmcdermott: This held out "20" is from the write call and can be suppressed if you set that call to a variable like "_" so it doesn't dirty doctest output.

mmcdermott (commented):
Can you show an example where the quantiles chosen are dynamic, end to end? E.g., a different full stage example to show how you can pick different quantiles?

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
mmcdermott: Dropped the `# noqa` in 5770267 by splitting the list-of-fields output over multiple lines (iterating with `for f in s.fields:`). Each line is now well under the 110-col limit, so the suppression was no longer needed.

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
mmcdermott: Replaced with a full `pprint` of the returned dict in 5770267. The doctest now covers both the base-schema case and the quantiles-present case end-to-end, showing the full structure rather than a single key lookup.

src/MEDS_transforms/stages/base.py
mmcdermott: Fixed in 5770267 by assigning the `write_text` call to `_` — the stray `20` byte-count no longer appears in the doctest output.

mmcdermott:
Addressed all four points in 5770267:

- **End-to-end example with non-default quantiles** — added `with_custom_quantiles` scenario under `src/MEDS_transforms/stages/aggregate_code_metadata/examples/`. Same static sample data as `on_raw_static_data`, but `cfg.yaml` declares `quantiles: [0.1, 0.9]` instead of the default `0.25/0.5/0.75`. The `out_metadata.yaml` expects a `values/quantile/0.1` + `values/quantile/0.9` struct, and the stage's dynamic schema produces exactly that. The test passes end-to-end via the existing `test_stage_scenario[aggregate_code_metadata-with_custom_quantiles]` parametrization — which would have failed under the old hardcoded 0.25/0.5/0.75 schema. README updated to describe the scenario.
- **`# noqa: E501`** — dropped by splitting the list-of-fields output over multiple lines with `for f in s.fields: print(f)` so each line fits comfortably under 110 cols.
- **Doctest on `aggregation_schema_updates`** — switched from key-access to full `pprint` of the returned dict, covering both the base-schema case and the quantiles-present case.
- **Stray `20` in `_resolve_output_schema_updates` doctest** — suppressed by binding `write_text` to `_`.

Full doctest suite: 94 passed. Registered-stage suite for `aggregate_code_metadata`: 6 passed (including the new scenario). Re-requesting Copilot.

src/MEDS_transforms/stages/aggregate_code_metadata/examples/with_custom_quantiles/cfg.yaml
```diff
@@ -0,0 +1,15 @@
+aggregations:
+  - code/n_occurrences
```
mmcdermott: for with custom quantiles, you can probably just include the quantiles, no?

mmcdermott: Good call — trimmed in 152179b. `cfg.yaml` now lists only `values/quantiles` (plus `do_summarize_over_all_codes`), and `out_metadata.yaml` dropped the unused aggregation columns. The scenario's intent — custom `probs` flowing through the dynamic schema — is now the only thing under test. Registered-stage test still passes.

src/MEDS_transforms/stages/base.py
```diff
@@ -721,6 +727,24 @@ def default_config(self, default_config: dict[str, Any] | DictConfig | Path | st
...
+        if self.output_schema_updates is None:
+            return {}
+        if not callable(self.output_schema_updates):
+            return dict(self.output_schema_updates)
+        stage_cfg: dict | None = None
+        if example_dir is not None:
+            stage_cfg_fp = example_dir / "cfg.yaml"
+            if stage_cfg_fp.is_file():
+                stage_cfg = OmegaConf.to_container(OmegaConf.load(stage_cfg_fp))
+        return dict(self.output_schema_updates(stage_cfg))
```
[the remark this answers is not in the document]
mmcdermott: Addressed in 2b4baa5 (earlier round): `_resolve_output_schema_updates` now falls back to `self.default_config` (via `OmegaConf.to_container`) when neither an explicit `stage_cfg` nor an `example_dir/cfg.yaml` is provided. As of 7cdf894 the resolver also deep-merges `default_config` with the overrides when both exist, so `Stage.__str__` / docgen see the same effective config the stage would get at runtime.

src/MEDS_transforms/stages/base.py
```diff
@@ -737,11 +761,12 @@ def test_cases(self) -> dict[str, StageExample]:
 
             if self.example_class.is_example_dir(example_dir):
                 scenario_name = example_dir.relative_to(self.examples_dir).as_posix()
+                schema_updates = self._resolve_output_schema_updates(example_dir)
                 test_cases[scenario_name] = self.example_class.from_dir(
                     stage_name=self.stage_name,
                     scenario_name=scenario_name,
                     example_dir=example_dir,
-                    **self.output_schema_updates,
+                    **schema_updates,
```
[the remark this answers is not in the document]
mmcdermott: Addressed in 2b4baa5 (earlier round). `Stage.test_cases` now parses each example's `cfg.yaml` once and passes the dict to both `_resolve_output_schema_updates` and `StageExample.from_dir` via a new keyword-only `stage_cfg=` arg on each. Single read per example.

src/MEDS_transforms/stages/base.py
```diff
@@ -721,6 +727,72 @@ def default_config(self, default_config: dict[str, Any] | DictConfig | Path | st
...
+            {'probs': 'from-default'}
+        """
+        if not callable(self.output_schema_updates):
+            return dict(self.output_schema_updates)
+
+        resolved_cfg: dict | None = stage_cfg
+        if resolved_cfg is None and example_dir is not None:
+            stage_cfg_fp = example_dir / "cfg.yaml"
+            if stage_cfg_fp.is_file():
+                resolved_cfg = OmegaConf.to_container(OmegaConf.load(stage_cfg_fp))
```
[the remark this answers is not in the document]
mmcdermott: Addressed in 7cdf894. `_resolve_output_schema_updates` now deep-merges `self.default_config` with the example / arg overrides via `OmegaConf.merge` before invoking the callable, so the schema function sees the same effective config it would see at pipeline runtime (overrides take precedence on conflict; default-only keys survive). New doctest shows a `default_config={"probs": ..., "n_bins": 8}` where the `n_bins` key carries through even when only `probs` is overridden.

src/MEDS_transforms/stages/aggregate_code_metadata/examples/README.md
```diff
@@ -1,7 +1,11 @@
 These examples show `aggregate_code_metadata` in several contexts:
 
 - **on_raw_static_data** and **with_not_split_defined_shards** -- comprehensive aggregation with
-    quantiles, counts, and distribution statistics on the full static sample data.
+    the default quantiles (`0.25 / 0.5 / 0.75`), counts, and distribution statistics on the full
+    static sample data.
+- **with_custom_quantiles** -- same data but with user-chosen `quantiles: [0.1, 0.9]`. The
+    `values/quantiles` struct schema is generated from the stage config, so non-default `probs`
+    round-trip through example validation (see issue #342).
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 7cdf894 — README now says "non-default `quantiles`" to match the config key name. Applied the suggestion verbatim.

src/MEDS_transforms/stages/base.py
```diff
@@ -476,7 +476,9 @@ class Stage:
     read_fn: READ_FN_T | None = None
     write_fn: WRITE_FN_T | None = None
```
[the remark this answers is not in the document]
mmcdermott: Addressed in 7cdf894. The Stage class-level Attributes docstring now documents both forms: "Either a plain dict, or a callable `(stage_cfg) -> dict` that derives the overrides from the stage's effective config (`default_config` deep-merged with the example/pipeline overrides). Use the callable form when the schema depends on user-chosen config, e.g. quantile probabilities."

src/MEDS_transforms/stages/base.py
```diff
@@ -721,6 +730,85 @@ def default_config(self, default_config: dict[str, Any] | DictConfig | Path | st
...
+        if override_cfg is None and example_dir is not None:
+            stage_cfg_fp = example_dir / "cfg.yaml"
+            if stage_cfg_fp.is_file():
+                override_cfg = OmegaConf.to_container(OmegaConf.load(stage_cfg_fp))
+
+        default_cfg = (
+            OmegaConf.to_container(self.default_config, resolve=False) if self.default_config else None
+        )
+
+        if default_cfg and override_cfg:
```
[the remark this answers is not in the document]
mmcdermott: Good catch — fixed in d59dc57. Switched from `if default_cfg and override_cfg:` to `if default_cfg is not None and override_cfg is not None:`, so an empty `cfg.yaml` / explicit `stage_cfg={}` no longer trips the truthiness trap. Defaults now survive the merge. Added a doctest at base.py:790 to lock it in.

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
```diff
@@ -808,6 +801,73 @@ def reducer(*dfs: Sequence[pl.LazyFrame]) -> pl.LazyFrame:
...
+            return pl.Struct({f"values/quantile/{q}": pl.Float32 for q in qs})
+    return None
+
+
+def aggregation_schema_updates(stage_cfg: dict | None = None) -> dict[str, pl.DataType]:
+    """Compose the example-time output schema overrides for this stage, including dynamic quantiles.
+
+    Static column types (``code/n_occurrences``, ``values/sum``, etc.) are always emitted. The
+    ``values/quantiles`` struct is generated from the example's ``aggregations`` config so that
+    non-default ``probs`` round-trip correctly (see issue #342).
```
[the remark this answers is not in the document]
mmcdermott: Fixed in d59dc57. Docstring now reads: "non-default `quantiles` (i.e. the quantile probabilities) round-trip correctly". Matches the user-facing config key.

PR #378, opened 2026-04-17
Regression test for reshard_to_split single-stage mode (closes #321)

### LINKED ISSUE #321: reshard_to_split doesn't work naturally in single-stage mode.
This is an abnormal use case, so maybe this is to be expected.

mmcdermott:
## Summary

Closes #321. Verification that ``reshard_to_split`` runs correctly in single-stage mode, plus a regression test to prevent future breakage.

When I tried to reproduce the reported failure on current ``dev`` I could not. The stage runs cleanly whether invoked as the only stage in a pipeline (output → ``output_dir/data``) or as a non-final stage (output → ``output_dir/reshard_to_split``). My guess is the underlying issue was fixed incidentally by the mapreduce/test-infrastructure refactors (#275, #290) since the stub was filed.

This PR just adds a regression test covering:
- single-stage invocation via ``MEDS_transform-stage``
- multi-split input (train / tuning / held_out)
- verification of ``.shards.json`` contents
- verification of each produced parquet's contents

## Test plan

- [x] ``uv run pytest tests/test_reshard_single_stage.py``: passed.

Closes #321

### DIFF
--- a/tests/test_reshard_single_stage.py
+++ b/tests/test_reshard_single_stage.py
@@ -24,19 +24,19 @@ def _write_dataset(input_dir: Path) -> None:
     base_time = datetime(2020, 1, 1, tzinfo=UTC)
 
     def _write(path: Path, subjects: list[int]) -> None:
-        # Use a real timestamp column so reshard_to_split's sort by DataSchema.time_name
-        # hits a well-typed column, matching production MEDS data.
+        # Use real timestamps with the MEDS schema's time dtype so reshard_to_split's sort by
+        # DataSchema.time_name hits a well-typed column that matches production MEDS data.
         rows = [
             {
-                "subject_id": s,
-                "time": base_time.replace(day=1 + i),
-                "code": "A",
-                "numeric_value": None,
+                DataSchema.subject_id_name: s,
+                DataSchema.time_name: base_time.replace(day=1 + i),
+                DataSchema.code_name: "A",
+                DataSchema.numeric_value_name: None,
             }
             for s in subjects
             for i in range(2)
         ]
-        (pl.DataFrame(rows, schema_overrides={DataSchema.time_name: pl.Datetime("us")}).write_parquet(path))
+        pl.DataFrame(rows, schema_overrides={DataSchema.time_name: pl.Datetime("us")}).write_parquet(path)
 
     _write(input_dir / "data" / "train" / "0.parquet", [1, 2, 3, 4])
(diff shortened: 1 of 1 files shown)


mmcdermott:
Both comments addressed in 1cf6ca1:

- **time column dtype**: switched to real tz-aware datetimes with an explicit pl.Datetime('us') schema override so reshard_to_split's sort by DataSchema.time_name hits a well-typed column matching production MEDS data.
- **ordering of subject IDs**: switched to set comparisons; kept a length assertion on the 4-subject train split so the 2+2 balance is verified without pinning an implementation-detail order.

mmcdermott:
All three round-2 comments addressed in 3d68e1b:

- **Hard-coded MEDS column names in row dicts**: switched to `DataSchema.subject_id_name`, `DataSchema.time_name`, `DataSchema.code_name`, `DataSchema.numeric_value_name` for the fixture; same for the splits `subject_id` key.
- **Hard-coded `subject_id` in final assertions**: switched to `DataSchema.subject_id_name`.
- **Redundant outer parens around the write_parquet call**: dropped.

mmcdermott:
Is this necessary? Do we not have a stage example of this stage in action that is run as a singleton through the test all registered stages fixture logic? I don't think we need a dedicated regression test here -- and if we do, it should be done through the dedicated example workflow.

mmcdermott:
You're right — this is redundant. The existing dedicated stage example at \`src/MEDS_transforms/stages/reshard_to_split/examples/\` is exercised end-to-end via \`test_stage_scenario[reshard_to_split-.]\` in \`tests/test_registered_stages.py\`, which:

- Runs \`MEDS_transform-stage <cfg> reshard_to_split input_dir=... output_dir=...\` (single-stage invocation — no outer pipeline runner).
- Uses the full \`SIMPLE_STATIC_SHARDED_BY_SPLIT\` dataset with train / tuning / held_out.
- Validates the reshard output (2 subjects per shard → 4 train shards, 1 tuning shard, 1 held_out shard) via \`check_outputs\` on every shard's parquet content.

That's exactly what this PR's handwritten regression test was trying to cover. The dedicated-example workflow does it more thoroughly (all shards, all subjects, schema-checked) and cheaper (no fixture drift).

Closing this PR unmerged. Closing #321 as can't-reproduce on current \`dev\`.

mmcdermott:
Closing unmerged — the existing stage example already covers single-stage `reshard_to_split` via `test_stage_scenario[reshard_to_split-.]`. See thread for details.

Issue #324, opened 2025-05-06
Stages should all export flexible_schema objects describing their output.


mmcdermott:
This can then be tied into the Stage Example for ease of use.

PR #379, opened 2026-04-17
Mechanism: Stages declare flexible_schema input/output contracts (refs #324)

mmcdermott:
## Summary

Part 1 of #324. Introduces the mechanism by which stages declare their input and output schemas, and demonstrates it on ``aggregate_code_metadata``.

- ``Stage`` now accepts ``input_schema``, ``output_schema``, ``metadata_input_schema``, ``metadata_output_schema`` (all ``flexible_schema.Schema`` subclasses, all optional).
- ``Stage.register`` forwards these kwargs unchanged.
- ``Stage.declared_schemas`` returns the four roles as a dict for consumers (pipeline-load-time validation, downstream composer checks in #56).
- ``aggregate_code_metadata`` declares ``DataSchema`` on ``input_schema`` and ``CodeMetadataSchema`` on ``metadata_output_schema`` — a case where the schemas actually differ, so the declaration carries real information. (An earlier draft annotated ``filter_subjects`` with ``DataSchema`` on both sides, but that was just restating the default; reverted and replaced per review.)

## Scope

Deliberately minimal. Full stage-by-stage migration is a follow-up so each migration can land with its own targeted review and tests (every stage is doing slightly different shape transforms and the ``output_schema`` may need to be derived from the stage config for some of them). This PR is the enabling plumbing.

## Test plan

- [x] Full non-parallel suite: **145 passed**.
- [x] New ``declared_schemas`` doctest covers all four role slots.
- [x] Registered-stage suite for ``aggregate_code_metadata``: 7 passed, including the existing example scenarios.

Refs #324


src/MEDS_transforms/stages/filter_subjects/filter_subjects.py
```diff
@@ -201,7 +202,7 @@ def filter_subjects_by_num_events(df: pl.LazyFrame, min_events_per_subject: int)
     return df.filter(pl.col("time").n_unique().over("subject_id") >= min_events_per_subject)
 
 
-@Stage.register
+@Stage.register(input_schema=DataSchema, output_schema=DataSchema)
```
mmcdermott: don't do this. If you want a proof of concept, use a stage in the example pipeline that actually has different stages. We don't need to deviate from defaults here, so don't.

mmcdermott (commented):
Can this not go straight into dev, but into another branch that is for the broader feature set this is a part of? If all we're adding is this, I'm not sure I see the point.

src/MEDS_transforms/stages/filter_subjects/filter_subjects.py
mmcdermott: Agreed — reverted in 222e4ce. `filter_subjects` has input and output both `DataSchema`, so the declaration was just restating the default with no demonstrative value.

Moved the POC to `aggregate_code_metadata` instead: `input_schema=DataSchema` (MEDS event records in) and `metadata_output_schema=CodeMetadataSchema` (codes.parquet out). Those are genuinely different schemas, so the declaration actually carries information.

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
```diff
@@ -809,5 +809,9 @@ def reducer(*dfs: Sequence[pl.LazyFrame]) -> pl.LazyFrame:
 
 
 stage = Stage.register(
-    map_fn=mapper_fntr, reduce_fn=reducer_fntr, output_schema_updates=AGGREGATION_SCHEMA_UPDATES
+    map_fn=mapper_fntr,
+    reduce_fn=reducer_fntr,
+    output_schema_updates=AGGREGATION_SCHEMA_UPDATES,
+    input_schema=DataSchema,
+    metadata_output_schema=CodeMetadataSchema,
 )
```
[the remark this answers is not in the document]
mmcdermott: Fixed — updated the PR description to reflect what the diff actually does. The POC now lives on `aggregate_code_metadata` (`input_schema=DataSchema` / `metadata_output_schema=CodeMetadataSchema`, which are genuinely different); `filter_subjects` stays on the default. See commit 222e4ce and the updated PR body. Earlier draft bouncing off `filter_subjects` is acknowledged in the description.

mmcdermott:
## Deep-review notes

CI is green and the schema-declaration mechanism is backwards-compatible. Below are non-blocking findings.

### Doc gap

- **``Stage.register`` docstring lists accepted kwargs but omits the four new schema parameters** (``input_schema``, ``output_schema``, ``metadata_input_schema``, ``metadata_output_schema``). They flow through ``**kwargs`` so calls work, but the docstring is now misleading. One-line update to ``**Args**`` would close this.

### Test gaps (the more impactful ones)

- **No tests for ``declared_schemas`` property.** All four roles, individually and in combination, are unexercised. Adding a small parameterized test that constructs a minimal ``Stage`` with each combination of (``None``, populated) for the four roles, then asserts the returned dict's keys/values, would lock the contract in.
- **No integration test for the ``aggregate_code_metadata`` migration.** That stage now declares ``input_schema=DataSchema`` + ``metadata_output_schema=CodeMetadataSchema`` but no ``output_schema`` — which is the correct shape for a MAPREDUCE stage with separate data and metadata outputs. Worth one test: load the registered stage, read ``declared_schemas``, assert the keys present match the intent. (The mapreduce execution itself is already tested elsewhere.)

### Design / doc

- **Asymmetric schema semantics** (a MAPREDUCE stage typically declares ``input`` + ``metadata_output`` but not ``output``) is correct but isn't documented anywhere. A note in the ``declared_schemas`` property docstring explaining that not all four roles must be declared, with a one-line example for MAP vs MAPREDUCE, would help downstream subclassers.

### Verified clean

- All four new fields default to ``None`` (immutable). No mutable-default trap.
- Existing ``Stage(...)`` callers are unaffected — kwargs are optional with ``None`` defaults.
- No off-by-one in the ``declared_schemas`` dict construction.

The docstring update + a small ``declared_schemas`` test is enough to land. Want me to push those?

mmcdermott:
Pushed 7dfb0d6 addressing the review: `Stage.register` Args docstring now lists the four flexible-schema kwargs plus the previously-undocumented `is_metadata`/`example_class`; `declared_schemas` property docstring now spells out the MAP vs MAPREDUCE shape patterns; new `tests/test_declared_schemas.py` covers a parameterized truth-table over the four roles plus an end-to-end check pinning the `aggregate_code_metadata` migration.

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
```diff
@@ -1031,5 +1031,9 @@ def aggregation_schema_updates(stage_cfg: dict | None = None) -> dict[str, pl.Da
 
 
 stage = Stage.register(
-    map_fn=mapper_fntr, reduce_fn=reducer_fntr, output_schema_updates=aggregation_schema_updates
+    map_fn=mapper_fntr,
+    reduce_fn=reducer_fntr,
+    output_schema_updates=aggregation_schema_updates,
+    input_schema=DataSchema,
```
mmcdermott: shouldn't this and the next be the default?

src/MEDS_transforms/stages/base.py
```diff
@@ -482,6 +485,10 @@ class Stage:
     output_schema_updates: dict[str, pl.DataType] | Callable[[dict | None], dict[str, pl.DataType]] | None = (
         None
     )
+    input_schema: SchemaLike | None = None
```
mmcdermott: I think by default these should use MEDS defaults (DataSchema, CodeMetadataSchema)

src/MEDS_transforms/stages/base.py
```diff
@@ -482,6 +485,10 @@ class Stage:
...
+    output_schema: SchemaLike | None = None
```
mmcdermott: Also, is this in conflict with output_schema_updates? Why do we need this and output_schema?

tests/test_declared_schemas.py
```diff
@@ -0,0 +1,73 @@
...
+        (DataSchema, None, None, None),
+        (None, DataSchema, None, None),
+        (None, None, CodeMetadataSchema, None),
+        (None, None, None, CodeMetadataSchema),
+        (DataSchema, DataSchema, None, None),
+        (DataSchema, None, None, CodeMetadataSchema),
+        (DataSchema, DataSchema, CodeMetadataSchema, CodeMetadataSchema),
+    ],
+)
+def test_declared_schemas_returns_all_four_roles(
```
mmcdermott: This test seems useless -- the logic it is testing is just "assign this to a parameter." A few doctests will catch this case, no?

src/MEDS_transforms/stages/aggregate_code_metadata/aggregate_code_metadata.py
mmcdermott: Fixed in bbdabb5 — `input_schema` and `metadata_output_schema` are now the defaults on `Stage`, so this registration drops both. The only remaining override is `output_schema=None` because this MAPREDUCE writes no data shards (only `metadata/codes.parquet`).

src/MEDS_transforms/stages/base.py
mmcdermott: Done in bbdabb5 — `input_schema`/`output_schema` now default to `DataSchema` and `metadata_input_schema`/`metadata_output_schema` default to `CodeMetadataSchema`. Stages that do not read/write a particular role set it to `None` explicitly.

src/MEDS_transforms/stages/base.py
mmcdermott: Not in conflict — they compose. `output_schema` is the *base* schema this stage writes (e.g., `DataSchema`), and `output_schema_updates` is a column-level diff layered on top (e.g., `{"new_col": pl.Int64}`). A downstream validator can compute the effective output schema as `output_schema + output_schema_updates`. Most stages now leave `output_schema` at its `DataSchema` default and just declare the diff via `output_schema_updates`. I added a note about this composition to the `declared_schemas` property docstring in bbdabb5.

tests/test_declared_schemas.py
mmcdermott: Agreed — dropped in bbdabb5. The property doctests now cover both the default-pickup case and the override-to-`None` case (the canonical MAPREDUCE-into-metadata pattern).

Issue #387, opened 2026-04-22
Hoist Path-spec StageExample infrastructure into `MEDS_transforms.stages.examples`


mmcdermott:
## Motivation

Two derived packages — [meds-torch-data](https://github.com/mmcdermott/meds-torch-data/blob/main/src/meds_torchdata/preprocessing/_stage_example.py) ([PR #86](https://github.com/mmcdermott/meds-torch-data/pull/86)) and [MEDS_extract](https://github.com/mmcdermott/MEDS_extract/pull/80) — have independently arrived at the same `StageExample` subclass pattern because the upstream class assumes outputs are either a parsed `MEDSDataset` under `data/*.parquet` or a single `metadata/codes.parquet`. Neither fits stages that emit:

- raw row-chunk parquets (`shard_events`: `data/<prefix>/[start-end).parquet`)
- JSON files (`split_and_shard_subjects`: `metadata/.shards.json`)
- per-subject schema / event-seq parquets with list-of-list columns (MTD's `tokenization`)
- NRT tensor-dict files (MTD's `tensorization`)
- intermediate parquet files whose schemas aren't MEDS-format

The workaround in both packages: declare inputs/outputs as `pathlib.Path` references to `yaml_to_disk` specs, subclass `StageExample`, override `is_example_dir` / `from_dir` / `check_outputs`, and add a per-suffix `_compare(expected_fp, actual_fp, rel, df_check_kwargs)` dispatcher.

## Concrete overlap

Substantive LOC (excluding comments/blanks): **MEDS_extract 220**, **MTD 149**.

| Concern | Near-identical in both |
|---|---|
| `want_data: Path \| None = None` (and `want_metadata: Path \| None` in MEDS_extract) field redeclaration | ✓ |
| `is_example_dir` checks for `out_data.yaml` (or `out_metadata.yaml`) | ✓ |
| `from_dir` resolves `in.yaml` / `out_data.yaml` / `cfg.yaml` to Paths, loads cfg as dict | ✓ |
| `check_outputs` structure: `yaml_disk(want_fp)` → walk expected → walk actual → assert set match → per-file compare | ✓ |
| Top-dir scoping: scan only top-level dirs mentioned in `expected` to avoid picking up hydra logs | ✓ |
| `_compare()` dispatcher keyed off `expected_fp.suffix`, with `.parquet` → `polars.testing.assert_frame_equal` | ✓ |

**What diverges** (and should stay package-specific):

[...]

PR #389, opened 2026-04-22
Fix render_content blowup when want_metadata is a Path (refs #387)

mmcdermott:
## Summary

Refs #387. Standalone prerequisite to the larger Path-spec `StageExample` hoist discussed in that issue.

## The bug

`StageExample.render_content` had MEDSDataset/Path dispatch for both `in_data` and `want_data`, but `want_metadata` fell straight through to `df_to_markdown(self.want_metadata)`. Downstream subclasses (e.g., [meds-torch-data](https://github.com/mmcdermott/meds-torch-data) and [MEDS-Extract](https://github.com/mmcdermott/MEDS_extract)) override the field to accept a `pathlib.Path` pointing at a `yaml_to_disk` spec; during doc generation that crashed with:

```
AttributeError: 'PosixPath' object has no attribute 'height'
```

## The fix

Mirror the `want_data` handling: when `want_metadata` is a `Path`, emit the raw YAML under an `**Expected output metadata files:**` heading; otherwise keep the existing DataFrame→Markdown-table rendering.

## What's not here

Deliberately minimal — no type-annotation widening, no new field plumbing, no comparator registry. Those land in the follow-up PR that implements Option B from #387.

## Test plan

- [x] New doctest in `render_content` exercising the `want_metadata: Path` branch.
- [x] Full non-parallel suite: **145 passed**, 2 pre-existing parallel-only failures unrelated.
- [x] Existing `want_metadata: DataFrame` doctest still passes (no change to that branch).

Refs #387


mmcdermott:
## Deep-review notes

Did a fresh review of the rendered changes. CI is green; below are non-blocking findings I traced but didn't act on.

### Verified false alarm

- **Doctest in ``runner.py`` for "Paths with spaces"**: I worried the dict-mutation step further down the doctest (``stage_runners["reorder_measurements"]["parallelize"] = ...``) would corrupt the new test's setup — but the new test uses ``reshard_to_split`` and runs *before* the mutation step, and CI is green. Not an issue.

### Open notes (low impact)

- **``render_content`` overrides are all-or-nothing**, not extensible per-field. A subclass that overrides ``render_content`` to handle (e.g.) JSON ``want_data`` doesn't inherit the base class's ``**Expected output files:**`` heading for any other case it falls through on — it owns the entire rendering. Worth a one-line note in the method docstring so subclassers know they can't ``super().render_content(...)`` and append.
- **``_safe_relative_to`` resolves symlinks** via ``Path.resolve()``, which is correct for canonical edit-link paths but worth mentioning in the docstring (currently silent on symlink behavior).
- **Scope**: ``shlex.quote`` (path-with-spaces fix) and ``_safe_relative_to`` are both legitimate improvements but tangential to the ``want_metadata: Path`` blowup. Not a blocker — the bundle is small and consistent with #387's surface area.

### TestGap (low value)

- ``test_docgen_renders_custom_example_class`` is happy-path only. Adding cases for empty ``want_data`` spec and malformed JSON would harden the extension hook, but the failure modes are obvious from the call graph.

Nothing here blocks merge.

mmcdermott:
Pushed 0e73fd8 with the two docstring nits: `render_content` now documents the all-or-nothing override semantics, and `_safe_relative_to` notes the symlink-dereference behavior of `Path.resolve()`.

PR #390, opened 2026-04-22
Hoist Path-spec StageExample infrastructure (issue #387, Option B)

### LINKED ISSUE #387: Hoist Path-spec StageExample infrastructure into `MEDS_transforms.stages.examples`

mmcdermott:
## Summary

Closes #387 (Option B). The upstream `StageExample` assumed outputs were either a `MEDSDataset` under `data/*.parquet` or a single `metadata/codes.parquet` — so two derived packages (meds-torch-data and MEDS-extract) independently reinvented the same ~200-LOC subclass: Path-based `want_data` / `want_metadata` plus a per-suffix comparator dispatcher.

This PR hoists that scaffolding into the base class so derived packages collapse to ~20 LOC.

## Depends on

**#389** — the `render_content` / `want_metadata: Path` bug-fix prerequisite. That PR is deliberately minimal (rendering dispatch only); this one adds the type widening + check_outputs dispatch + comparator registry on top.

### Type widening
- `want_data: MEDSDataset | Path | None`
- `want_metadata: pl.DataFrame | Path | None`

A `Path` is interpreted as a `yaml_to_disk` spec describing the expected output tree.

### `from_dir` graceful fallback
- Try `MEDSDataset.from_yaml(out_data_fp)` / `read_metadata_only(out_metadata_fp)` first. On `ValueError` / `TypeError`, store the `Path` itself and let `check_outputs` dispatch to the file-spec path later. Mirrors the existing `in_data` pattern exactly.

### New `check_outputs` dispatch
- `MEDSDataset` / `pl.DataFrame` branches are byte-identical to before (no regression for existing stages).
- `Path` branches go through the new `_check_path_spec`, which:
  1. materializes the spec into a tempdir via `yaml_to_disk.yaml_disk`;
  2. walks the expected tree, compares each file to its counterpart under the actual output dir via a per-suffix `ComparatorFn`;
  3. asserts the actual tree has no unexpected files (unless `tolerate_unexpected=True`), after filtering out `skip_dirs` / `skip_files`.

### Policy fields (dataclass fields, not ClassVars)
- `skip_dirs: frozenset[str]` — defaults to `{.logs, .hydra}` (the usual Hydra / logging byproducts both downstream packages reinvented).
- `skip_files: frozenset[str]` — empty default.
- `tolerate_unexpected: bool` — defaults to strict.
- `suffix_comparators: dict[str, ComparatorFn] | None` — per-instance override of the class default.

### Comparator registry (no process-global state)
- Class-level default: `SUFFIX_COMPARATORS: ClassVar[dict[str, ComparatorFn]] = {".parquet": _compare_parquet}`.
- Subclasses register additional comparators via standard class-var override (`class MTDStageExample(StageExample): SUFFIX_COMPARATORS = {..., ".nrt": _compare_nrt}`).
- Per-instance `suffix_comparators` overrides the class default for one-offs (e.g. relaxed tolerance on a specific example).
- No `@register_comparator` decorator — explicitly avoided per the design-discussion comment.

### Comparator signature
```python
ComparatorFn = Callable[[Path, Path, CompareContext], None]
```
`CompareContext` is a frozen dataclass carrying `rel: Path` (for error messages) and `tolerances: Mapping[str, Any]` keyed by suffix. `df_check_kwargs` is plumbed through as `ctx.tol_for('.parquet')` — so adding a new format doesn't require signature changes.

### XOR relaxation
The `want_data` XOR `want_metadata` rule still fires when both are `MEDSDataset` + `pl.DataFrame` (the original ambiguity). It's relaxed when either is a `Path` — extraction stages legitimately describe both data and metadata file trees simultaneously.

## What this doesn't do

- No changes to `example_class` — stages that need fully custom rendering (`JsonOutputStageExample`-style) keep their escape hatch.
- No changes to `render_content` — #389 handled the `want_metadata: Path` rendering bug on its own.
- No downstream migrations — those land as PRs against the MTD and MEDS-extract repos once this merges.

## Test plan

- [x] 18 new tests in `tests/test_filespec_stageexample.py`:
  - Path fallback in `from_dir` + existing MEDSDataset parsing unchanged
  - `check_outputs` dispatch for Path spec (match / mismatch / missing-file / unknown-suffix)
  - Policy fields (`tolerate_unexpected`, `skip_dirs`, `skip_files`)
  - Class-level comparator registration
  - Per-instance `suffix_comparators` wins over class default
  - `_compare_parquet` tolerance bridging
  - MEDSDataset legacy branch un-regressed
- [x] Full test suite: **177 passed**, 2 pre-existing parallel-only failures unrelated.
- [x] All existing doctests still pass.

Refs #387

### DIFF
--- a/src/MEDS_transforms/stages/examples.py
+++ b/src/MEDS_transforms/stages/examples.py
@@ -9,10 +9,11 @@ from __future__ import annotations
 import subprocess
 import tempfile
 import textwrap
+from collections.abc import Callable, Mapping
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, ClassVar
+from typing import Any, ClassVar
 
 import polars as pl
 from meds import DatasetMetadataSchema, code_metadata_filepath
@@ -22,9 +23,6 @@ from omegaconf import DictConfig, OmegaConf
 from polars.testing import assert_frame_equal
 from pretty_print_directory import list_directory
 from yaml import load as load_yaml
-
-if TYPE_CHECKING:
-    from collections.abc import Callable
 
 _SPACE = "    "
 _BRANCH = "│   "
@@ -275,6 +273,54 @@ class TestEnv:
             lines.extend(textwrap.indent(cfg_yaml_contents, _SPACE + _BRANCH).splitlines())
         lines.append(f"  - Script: {self.script}")
         return "\n".join(lines)
+
+
+@dataclass(frozen=True)
+class CompareContext:
+    """Context bundle passed to per-suffix output comparators.
+
+    Kept minimal on purpose: comparators that need new tolerance knobs (e.g. NRT atol/rtol) pull
+    them from ``tolerances`` under their own suffix key, so the comparator signature never has
+    to grow.
+
+    Attributes:
+        rel: The expected file's path relative to the expected root. Used in error messages so
+            the failure points at the logical location a stage author cares about.
+        tolerances: Per-suffix tolerance bundles. A comparator looks up ``tolerances.get(".parquet")``,
+            ``tolerances.get(".nrt")``, etc. Empty dict if the caller didn't configure any.
+    """
+
+    rel: Path
+    tolerances: Mapping[str, Any] = field(default_factory=dict)
+
+    def tol_for(self, suffix: str) -> dict:
+        """Return a (possibly empty) tolerance dict for ``suffix``.
+
+        Always a fresh dict.
+        """
+        return dict(self.tolerances.get(suffix, {}))
+
+
+#: Comparator signature. Must ``raise AssertionError`` on mismatch; return value is ignored.
+ComparatorFn = Callable[[Path, Path, "CompareContext"], None]
+
+
+def _compare_parquet(expected_fp: Path, actual_fp: Path, ctx: CompareContext) -> None:
+    """Default comparator for ``.parquet`` outputs.
+
+    Uses ``pl.read_parquet(..., glob=False)`` on both sides (raw shard filenames like
+    ``[0-10).parquet`` that appear in extraction pipelines carry glob metacharacters). Tolerance
+    comes from ``ctx.tol_for('.parquet')`` and is forwarded to ``polars.testing.assert_frame_equal``.
+    """
+    try:
+        want = pl.read_parquet(expected_fp, glob=False)
+        got = pl.read_parquet(actual_fp, glob=False)
+    except Exception as e:
+        raise AssertionError(f"Failed to read parquet at {ctx.rel}: {e}") from e
+    try:
+        assert_frame_equal(want, got, **ctx.tol_for(".parquet"))
+    except AssertionError as e:
+        raise AssertionError(f"Parquet mismatch at {ctx.rel}:\n  want:\n{want}\n  got:\n{got}") from e
 
 
 @dataclass
@@ -848,19 +894,44 @@ class StageExample:
     stage_name: str
     scenario_name: str | None = None
     stage_cfg: dict = field(default_factory=dict)
-    want_data: MEDSDataset | None = None
-    want_metadata: pl.DataFrame | None = None
+    want_data: MEDSDataset | Path | None = None
+    want_metadata: pl.DataFrame | Path | None = None
     in_data: MEDSDataset | Path | None = None
     pipeline_cfg: dict = field(default_factory=dict)
     do_use_config_yaml: bool = False
     df_check_kwargs: dict | None = None
+    #: Top-level directory names to ignore when comparing a Path-spec output tree against the
+    #: actual output directory. Defaults cover the usual Hydra / logging byproducts.
+    skip_dirs: frozenset[str] = field(default_factory=lambda: frozenset({".logs", ".hydra"}))
+    #: File names (matched anywhere in the actual output) to ignore when reconciling expected
+    #: vs. actual. Empty by default; subclasses set this to ignore stage-specific auxiliary files.
+    skip_files: frozenset[str] = field(default_factory=frozenset)
+    #: If ``True``, actual-output files that the Path spec does not mention are silently ignored
+    #: rather than raising. Defaults to strict behavior.
+    tolerate_unexpected: bool = False
+    #: Per-instance override of the per-suffix comparator map. ``None`` means "use the
+    #: subclass's ``SUFFIX_COMPARATORS`` class var (merged with base defaults)".
+    suffix_comparators: dict[str, ComparatorFn] | None = None
 
     BASE_PACKAGE: ClassVar[str] = "__null__"
+    #: Default per-suffix comparators, resolved at the class level. Downstream packages subclass
+    #: ``StageExample`` and override this to register format-specific comparators (MTD's ``.nrt``,
+    #: MEDS-extract's ``.json``/``.yaml``). Per-instance overrides via ``suffix_comparators``
+    #: take precedence over this.
+    SUFFIX_COMPARATORS: ClassVar[dict[str, ComparatorFn]] = {".parquet": _compare_parquet}
 
     def __post_init__(self):
         if self.want_data is None and self.want_metadata is None:
             raise ValueError("Either want_data or want_metadata must be provided.")
-        elif self.want_data is not None and self.want_metadata is not None:
+        elif (
+            self.want_data is not None
+            and self.want_metadata is not None
+            and not isinstance(self.want_data, Path)
+            and not isinstance(self.want_metadata, Path)
+        ):
+            # The XOR rule is about MEDSDataset + DataFrame ambiguity; Path-spec examples may
+            # legitimately describe both data and metadata file trees simultaneously (e.g.
+            # extraction stages emit raw shard parquets AND a metadata/codes.parquet in one run).
             raise ValueError("Either want_data or want_metadata must be provided, but not both.")
 
         if self.scenario_name == ".":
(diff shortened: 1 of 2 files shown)


mmcdermott:
## Deep-review notes

CI is green and the type widening is backwards-compatible. Below are findings to consider; flagging the most important first.

### BUG (small, should probably address) — ``from_dir`` graceful fallback may mask YAML syntax errors

``StageExample.from_dir`` falls back to storing the raw ``Path`` on either ``ValueError`` *or* ``TypeError`` from ``MEDSDataset.from_yaml`` / ``read_metadata_only``. ``ValueError`` covers the intended "schema mismatch — treat as a yaml_to_disk spec" path, but ``TypeError`` is broader and could hide real issues (e.g. corrupted YAML where parsing yields the wrong types entirely). Suggestion: narrow the catch to just ``ValueError``, or add a debug-level log before swallowing so a downstream ``check_outputs`` failure can be cross-referenced.

### TEST GAPS

- **No test for ``want_data`` AND ``want_metadata`` simultaneously as Paths** when ``check_outputs`` is actually invoked — only field storage is asserted. The XOR-relaxation for the Path/Path case is the most novel behavior in this PR; would be reassuring to see it exercised end-to-end.
- **Comparator error messages don't assert ``ctx.rel`` is included.** ``test_check_outputs_path_spec_mismatch_propagates_comparator_error`` checks that the error bubbles, but if a comparator raises without including the relative path, the user has no way to know which file failed. One-line addition: ``assert "<rel-path>" in str(excinfo.value)``.
- **``skip_files`` only tested at the root.** The check uses ``fp.name``, which is correct, but a nested test (e.g., ``data/sub/README``) would lock that in.

### DESIGN (style)

- ``frozenset({".logs", ".hydra"})`` and ``frozenset()`` as ``default_factory=lambda: ...`` could be ``default_factory=frozenset`` plus a ``default=...`` for the populated case. Minor.

### Verified clean

- Type widening (``MEDSDataset | Path``) is backwards-compatible; XOR rule still fires for the original ``MEDSDataset + DataFrame`` ambiguity, only relaxed when at least one side is a Path.
- Per-instance ``suffix_comparators`` correctly beats the class-level ``SUFFIX_COMPARATORS``.
- Tempdir cleanup on ``check_outputs`` assertion failure is correct (the ``with`` is outside the comparator loop).
- Frozen-dataclass ``CompareContext`` mutability — the ``tolerances`` field is typed as ``Mapping`` (read-only protocol); ``tol_for`` returns a fresh dict.

Recommend addressing the ``ValueError | TypeError`` catch and adding the rel-path assertion; the rest are stylistic.

mmcdermott:
Pushed 231b0b2 addressing the review notes: narrowed `from_dir` exception catch to `ValueError` only, added the `ctx.rel` assertion to the comparator-error test, added the nested-dirs `skip_files` test, and added the dual-Path `check_outputs` test. The dual-Path test surfaced a real bug — `_check_path_spec` ran its no-unexpected-files check per-spec, so each spec rejected the other's files as unexpected. Refactored: split out `_compare_path_spec` (returns expected rel-set), kept `_check_path_spec` as the single-spec wrapper, and made `check_outputs` run a union-aware unexpected check when both fields are Paths.

Issue #182, opened 2024-08-27
Duplication between `text_value` and `numeric_value` should be ignored when possible.


mmcdermott:
This is likely a good thing to consider adding to the default extract pipeline, as it is not uncommon for source datasets to have duplication here.

Tagging @EthanSteinberg for tracking

PR #394, opened 2026-04-25
Add `deduplicate_events` stage (closes #393)

### LINKED ISSUE #393: Add a generic `deduplicate_events` stage
## Problem

There's no general-purpose stage today for "drop duplicate event rows". Existing options are partial:

- `MEDS_extract.convert_to_MEDS_events` has a `do_dedup_text_and_numeric` flag, but it's per-extracted-table (not a generic post-pipeline stage) and only handles the text↔numeric conflict from #182.
- `MEDS_extract.merge_to_MEDS_cohort.merge_subdirs_and_sort` takes a `unique_by` kwarg, but it's an internal helper inside a stage that does several other things — not callable on its own and not configurable by the user.
- `filter_measurements`, `aggregate_code_metadata`, `fit_vocabulary_indices` cover related concerns (filtering, aggregation, uniqueness checks) but none drop duplicate event rows.

## Use case

Multiple downstream consumers want this. Examples:

1. **MIMIC-IV `deathtime`** ([MEDS_extract#65](https://github.com/mmcdermott/MEDS_extract/issues/65)): `hosp/admissions` records every admission's recorded `deathtime`; collapsing them per-subject is a clinical/quality decision and ETL authors currently do it ad-hoc in `pre_MEDS.py`. With this stage they'd extract every recorded deathtime as its own `MEDS_DEATH` event and reduce post-extraction.
2. **Heterogeneous `text_value` / `numeric_value` duplicates** (#182): same `(subject_id, time, code)` recorded once with text + once with numeric. Today's `do_dedup_text_and_numeric` flag is the workaround.
3. **Re-entry of the same observation across overlapping source files**: HIRID's per-day chunks sometimes restate the previous day's last reading.

## Proposed shape

Mirror the `filter_measurements` layout — new `MEDS_transforms/stages/deduplicate_events/`:

```yaml
# stage_cfg defaults
by: [subject_id, time, code, numeric_value, text_value]   # uniqueness-key columns; defaults to all schema cols
on_conflict: error                                         # one of: error | first | last | drop
```

Semantics by group of rows that share `by` values:
- All other columns agree (true duplicate) → collapse to one row in **every** policy.
- Any other column disagrees:
  - `error` (default): raise with an example offending group. Conservative — surfaces a real data-quality issue rather than silently masking it.
  - `first` / `last`: keep that row in source order.
  - `drop`: remove **all** rows in the disagreement group.

Default `error` is intentionally strict because silent collapse of disagreeing rows is the failure mode that makes #182 / #65 hard to reason about.

## Acceptance

- New stage at `src/MEDS_transforms/stages/deduplicate_events/` with `__init__.py`, `deduplicate_events.py`, `config.yaml`, `examples/`.
- Registered as an entry point in `pyproject.toml` and re-exported from `stages/__init__.py`.
- Doctests on the stage covering: pure-duplicate collapse, each `on_conflict` policy, default-by inference, and the error-message shape (must name the offending group).
- One end-to-end example under `examples/simple/` matching the existing layout.

I'll send a PR.

Refs MEDS_extract#65, #182.

mmcdermott:
Closes #393. References mmcdermott/MEDS_extract#65, #182.

## Summary

Adds a generic, configurable post-pipeline stage for "drop duplicate event rows". Today the closest existing options are:

- `MEDS_extract.convert_to_MEDS_events.do_dedup_text_and_numeric` — flag, per-extracted-table, only handles the text↔numeric conflict from #182.
- `MEDS_extract.merge_to_MEDS_cohort.merge_subdirs_and_sort.unique_by` — internal helper inside another stage, not directly callable.
- `filter_measurements`, `aggregate_code_metadata`, `fit_vocabulary_indices` — related concerns (filtering, aggregation, uniqueness checks) but none drop duplicate event rows.

## Behavior

Two rows are "in the same `by` group" when their `by`-column values match. For each such group:

- Every non-`by` column agrees (true duplicate) → collapse to one row in **every** policy.
- Any non-`by` column disagrees → behavior depends on `on_conflict`:
  - `error` (default): raise `DuplicateEventConflictError` naming the offending group. Conservative — surfaces a real data-quality problem rather than silently masking it.
  - `first` / `last`: keep that row in source order.
  - `drop`: remove **all** rows in the disagreement group.

Default `by` is the four mandatory MEDS event-row columns (`subject_id`, `time`, `code`, `numeric_value`). Pipeline configs that carry optional schema columns (e.g. `text_value`) should override `by` to include them; otherwise rows that share the default key but disagree on `text_value` are silently collapsed under `first`/`last`/`drop`. The default `error` policy still raises in that case.

## Use cases (referenced in the issue)

- **MIMIC-IV `deathtime`** ([MEDS_extract#65](https://github.com/mmcdermott/MEDS_extract/issues/65)): extract every recorded death-time as its own `MEDS_DEATH` event, then run `deduplicate_events: {on_conflict: first}` (or `error` if the dataset claims they should agree). Eliminates the bespoke `fix_static_data` pre-MEDS code.
- **Heterogeneous `text_value` / `numeric_value` duplicates** (#182).
- **HIRID per-day chunk overlap** — same observation restated across boundary days.

## Layout

Mirrors `filter_measurements`:

```
src/MEDS_transforms/stages/deduplicate_events/
  __init__.py
  config.yaml              # by: null, on_conflict: error
  deduplicate_events.py
  examples/
    __init__.py
    cfg.yaml               # on_conflict: first
    in.yaml                # 14 rows; mix of pure dups + by-key disagreement
    out_data.yaml          # 11 rows after dedup
    README.md
```

Registered as a `MEDS_transforms.stages` entry point in `pyproject.toml` and re-exported from `stages/__init__.py`.

## Test plan

- [x] Doctests in `deduplicate_events.py` cover: pure-duplicate collapse, each `on_conflict` policy (`error` / `first` / `last` / `drop`), default-by inference, custom-by override, missing-from-input column tolerance, and unknown-policy validation.
- [x] `tests/test_registered_stages.py::test_stage_help[deduplicate_events]` — passes (entry-point registration confirmed).
- [x] `tests/test_registered_stages.py::test_stage_scenario[deduplicate_events-.]` — passes (end-to-end via the example fixture).
- [x] Full test suite: 164 passed; the 3 failing tests (`test_pipeline.py::test_example_pipeline_parallel_*`, `example/README.md`) are pre-existing on `dev` (verified via `git stash` + re-run), unrelated to this PR.
- [x] `pre-commit run --files <changed>`: all hooks pass.


mmcdermott:
## Deep-review notes

CI is green; below are non-blocking findings.

### Open questions / would benefit from your call

- **Default ``on_conflict: error`` is a strong stance.** SQL-style implicit dedup is the default users coming from other ETL frameworks expect; ``error`` is fail-fast on the first conflict. This is a fine default for data-quality-first pipelines but worth a one-line comment in ``config.yaml`` (or the stage docstring's "Configuration:" block) explaining the philosophy. Want me to add that?
- **``output_schema_updates`` not declared.** This is a pure structural operation (rows drop, schema unchanged), so ``output_schema_updates={}`` is correct. But declaring it explicitly via ``@Stage.register(output_schema_updates={})`` makes the contract visible to schema-validation tooling and to readers skimming the file. Worth doing for forward-compat with #379 (flexible_schema)?

### Test gaps (low value but easy)

- No doctest covers an **empty input DataFrame** — the unique/group-by code path may have polars edge cases there.
- No coverage for **list/struct value columns** in the non-key columns. The current implementation operates on group-by, which handles list columns fine, but a regression test would lock it in.

### Performance (acceptable trade-off)

- The ``on_conflict="error"`` path collects the full lazy frame to materialize a detailed error message. For pipelines with many small disagreement groups this is wasteful, but the alternative — a lazy count-then-collect — would double the scan in the success case. Current trade-off favors clear failure diagnostics over fast-path performance, which is the right call for a data-cleaning stage. Documenting the trade-off in the docstring would help future maintainers.

Nothing here blocks merge. The two doc/declaration nits (``on_conflict`` default rationale, explicit empty ``output_schema_updates``) are the most actionable; let me know which you'd like and I'll push.

mmcdermott:
Pushed 02999da addressing the two doc/declaration nits: `@Stage.register(output_schema_updates={})` now declares the no-op-on-schema contract explicitly (forward-compat with #379), and the `on_conflict='error'` default rationale is documented in the docstring. Empty-input / list-column doctests left for a follow-up — the underlying polars semantics aren't surprising and the cost-benefit didn't justify the doctest noise.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+    """Rows sharing the same ``by`` columns disagree on a non-``by`` column.
+
+    Raised by :func:`deduplicate_events` under ``on_conflict="error"`` (the default).
+    The message names the offending ``by`` group and, when feasible, the disagreeing
+    column(s) and example values — diagnosing the data-quality issue is the whole
+    point of the strict default.
+    """
+
+
+@Stage.register(output_schema_updates={})
```
mmcdermott, line 74: Do you need to specify the output schema updates if it is empty?

src/MEDS_transforms/stages/deduplicate_events/examples/README.md
```diff
@@ -0,0 +1,9 @@
+# `deduplicate_events` example
+
+Demonstrates the `on_conflict: first` policy on a single shard with three kinds of duplicate-shaped rows:
+
+- Two identical `(239684, ?, EYE_COLOR//BROWN, ?)` rows — true duplicate, collapses to one.
```
mmcdermott: why are these "?"s? Should these be `null` instead or something?

src/MEDS_transforms/stages/deduplicate_events/config.yaml
```diff
@@ -0,0 +1,12 @@
+# Columns whose value-tuple defines uniqueness. Defaults to the four mandatory MEDS
```
mmcdermott: technically, only code, subject_id, and time are mandatory. both numeric_value and text_value are optionals. I think the default here should be to use all fields in the MEDS schema -- which you can get programmatically via https://github.com/Medical-Event-Data-Standard/meds/blob/main/src/meds/schema.py#L155

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from omegaconf import DictConfig
+
+
+# Default ``by`` set: the four mandatory MEDS event-row columns. ``text_value`` is
+# part of the schema but optional, so callers who want it as part of the uniqueness
+# key must list it explicitly via ``by``.
+_DEFAULT_BY: tuple[str, ...] = (
```
mmcdermott: If you make the default by of None include the whole dataschema, this will likely become reduendant.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+    DataSchema.subject_id_name,
+    DataSchema.time_name,
+    DataSchema.code_name,
+    DataSchema.numeric_value_name,
+)
+
+_VALID_POLICIES: frozenset[str] = frozenset({"error", "first", "last", "drop"})
```
mmcdermott: would a StrEnum be cleaner here? feel free to push back if not.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+class DuplicateEventConflictError(ValueError):
```
mmcdermott, line 64: Why do you want a separate error type here? Just explain in a comment on GH, don't modify the code or documentation.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+        before any data is touched:
+
+        >>> deduplicate_events(DictConfig({"on_conflict": "median"}))
+        Traceback (most recent call last):
+            ...
+        ValueError: deduplicate_events: unknown on_conflict 'median'. ...
+    """
+    on_conflict = stage_cfg.get("on_conflict", "error")
+    if on_conflict not in _VALID_POLICIES:
+        supported = ", ".join(sorted(_VALID_POLICIES))
```
mmcdermott: Are there other patterns like this (or once you use a StrEnum) in the codebase? If so, it might be worth DRYing them in a later PR, and if so you should file an issue pointing to all such patterns and suggesting resolutions.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+            return df.unique(subset=by_cols, keep="first", maintain_order=True)
+        if on_conflict == "last":
+            return df.unique(subset=by_cols, keep="last", maintain_order=True)
+
+        # ``error`` and ``drop`` both need to identify disagreement groups. Strategy:
+        # collapse exact duplicates first (they're fine in every policy), then count
+        # remaining rows per ``by`` group — anything left with count > 1 is a group
+        # whose rows agree on ``by`` but disagree on some other column.
+        deduped = df.unique(maintain_order=True)
+        if on_conflict == "drop":
```
mmcdermott: I'm not sure we should support a drop policy -- at the very list it should log extensively.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+            return deduped.filter(pl.len().over(by_cols) == 1)
+
+        # on_conflict == "error": materialize the conflicts so the message can name them.
```
mmcdermott, line 417: A `warn` option might be nice, where some information can be logged about the issues but the pipeline doesn't error? Or an independent option so that warn can be toggled with first and last? Additionally, it might be worth having an option where the rest of the columns get captured in a struct column or something -- e.g., so you add a struct column that has {n_collisions: int} then lists for all columns in the df not in the keys with all the colliding options?

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+        first_key = conflicts.select(by_cols).row(0, named=True)
+        offending = conflicts.filter(
+            pl.all_horizontal(
+                *[pl.col(c).is_null() if v is None else pl.col(c) == v for c, v in first_key.items()]
+            )
+        )
+        raise DuplicateEventConflictError(
+            f"deduplicate_events: {len(conflicts)} rows in {conflicts.select(by_cols).n_unique()} "
+            f"group(s) share their `by` columns ({by_cols!r}) but disagree on a non-`by` column. "
+            f"First offending group: by={first_key!r}, rows=\n{offending}\n"
```
mmcdermott, line 436: Before this, you might need to collect offending, right? I'm not sure either way.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
```diff
@@ -0,0 +1,290 @@
...
+        ...     "subject_id": [1, 1],
+        ...     "time": [None, None],
+        ...     "code": ["MEDS_DEATH", "MEDS_DEATH"],
+        ...     "numeric_value": [None, None],
+        ...     "text_value": ["2020-03-01", "2020-03-05"],   # disagrees!
+        ... }).lazy()
+        >>> deduplicate_events(DictConfig({}))(df_disagree).collect()
+        Traceback (most recent call last):
+            ...
+        MEDS_transforms.stages.deduplicate_events.deduplicate_events.DuplicateEventConflictError: ...
```
mmcdermott: Can you show more of the error message here? It's full contents should be deterministic as this is a doctest, and it would be useful for developers.

mmcdermott (commented):
You need some

src/MEDS_transforms/stages/deduplicate_events/examples/README.md
mmcdermott: Fixed in dd9f72c — the README now uses explicit `null` (and the example trigger guidance was rewritten now that the default `by` is the full DataSchema).

src/MEDS_transforms/stages/deduplicate_events/config.yaml
mmcdermott: Done in dd9f72c — `_default_by_cols()` reads `DataSchema.schema()` at runtime so the full field set (`subject_id`, `time`, `code`, `numeric_value`, `text_value`) is the default. Caveat about overriding for `text_value` removed accordingly. Schemas that omit `text_value` still dedupe correctly because `by` columns missing from the input are silently dropped from the key.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Comment removed in dd9f72c — once the default became the full DataSchema, the `text_value` caveat was wrong rather than just stale, so it had to go anyway.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Switched to `OnConflictPolicy(StrEnum)` in dd9f72c. Matches the existing pattern in this codebase (`StageType`, `MatchReviseMode`, `ColExprType`, `VocabularyOrdering`, `MetadataFn`). Same Hydra-string ergonomics, single source of truth, and the dispatch is now `policy is OnConflictPolicy.FIRST` instead of string-equality.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Reasoning, per your request: callers that want to handle this specific failure (e.g., write a quarantine shard, fall back to `drop`) can `except DuplicateEventConflictError` without also catching unrelated `ValueError`s the function raises (bad `on_conflict` value, bad config, etc.). The class IS a `ValueError` subclass, so existing `except ValueError` handlers continue to work — but new code that wants to be precise has a narrower target. No code/doc change.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Checked the codebase: the `_VALID_POLICIES = frozenset({...})` literal-string-set pattern only appears here. Other enum-shaped configs already use `StrEnum` (the five I listed in the reply above). With this PR also using `StrEnum`, there is nothing left to DRY — no follow-up issue needed.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Done in dd9f72c — `drop` now emits a `logger.warning` per call, naming the dropped `by` keys (truncated at 10 with the total row + group counts). Cost: an extra collect of the conflicting groups to materialize the list. Addressed alongside the next comment about a `warn` policy.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Holding off on this until you sign off — it's a real design question and I'd rather not guess. Current PR adds the `logger.warning` to `drop` so it isn't silent. For the broader proposal:

**Option A — `warn` as a fifth policy.** Behaves like `error` (collects offending groups, logs at WARNING with the same message body) but does NOT raise; downstream gets the deduped frame with conflict groups still present. Simple, pairs cleanly with the existing four policies.

**Option B — orthogonal `log: silent | warn` toggle.** Decouples "what to do" from "what to say". `(error, silent)` = today's strict default; `(drop, warn)` = today's drop with the new logger.warning; `(first/last, warn)` = log every collapse. More configurations, more cells in the test matrix.

**Option C — capture conflicts in a struct column.** Add an opt-in `capture_conflicts: bool` that adds a `_conflicts: struct{n_collisions, columns, values}` column to disagreement-group rows. Useful for downstream auditability, but breaks the `output_schema_updates={}` contract and complicates the pipeline schema.

My preference would be **Option A** for this PR (cheap, additive, doesn't touch the schema). Option B/C feel like follow-up issues. Want me to push Option A here, or split into a separate PR?

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Looking again — `conflicts` is already a collected DataFrame from the line `conflicts = (deduped...).collect()` a few lines above. So `conflicts.filter(...)` returns a DataFrame, not a LazyFrame; no further collect is needed. The `pl.col(c).is_null()` / `== v` expressions are valid on both, so the chain works either way; we just do not need a redundant collect.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Done in dd9f72c — the doctest now shows the full deterministic message including the `by` columns list, the offending key, the `shape: (2, 6)` table of disagreeing rows, and the closing `Fix the source data, ...` hint. `# doctest: +NORMALIZE_WHITESPACE` is added because the message spans multiple lines.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Try option C here so we can see how they looks. I think that will be useful for data ETL pipelines where this may be used.

mmcdermott: Also add option B though, too, as we may still want it.

Also there may be a capture warnings caplog based doctest helper somewhere in the repo you can use for better testing here.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: I meant the validation of an input parameter belonging to a StrEnum of config options and giving a good error message if not. That pattern

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Done in 236fdbd — `OnConflictPolicy.CAPTURE` keeps every row in a disagreement group and adds a `_conflicts` struct column with `{n_collisions, <col>_values for each non-by col}`. Rows outside any conflict group get `null`. New doctest exercises the struct shape and value lists.

Implementation note: the join back to the per-group aggregate uses `nulls_equal=True` — MEDS event rows commonly have null `time` (static facts) and null `text_value` (numeric measurements), and the default null-not-equal join semantics would silently miss those rows.

mmcdermott: Both done in 236fdbd:

- **Option B**: orthogonal `log_conflicts: bool = False` config key. When True, opts every non-`error` policy into the same WARNING-level offending-key log that `drop` uses unconditionally. `drop` keeps logging unconditionally (data loss is destructive). The materialize-conflicts-and-log block was DRYed into a small `_log_offending` helper shared by `drop` and the `log_conflicts` paths.
- **caplog helper**: yes, `conftest.py` exposes `print_warnings` (a caplog-based context manager) into the doctest namespace. Used it for the new `log_conflicts` doctest so the warning output is captured deterministically.

src/MEDS_transforms/stages/deduplicate_events/deduplicate_events.py
mmcdermott: Got it — found the same pattern in `fit_vocabulary_indices` (validates `ordering_method` against `VocabularyOrdering_METHODS.keys()`) and `match_revise` (validates `match_revise_mode` against `{x.value for x in MatchReviseMode}`). Same shape, slightly different error text in each.

Filed #396 with a proposed `coerce_strenum(raw, enum_cls)` helper and migration acceptance criteria for all three call sites. Out of scope for this PR (one-PR-one-purpose) but tracked so the next StrEnum-stage author doesn't reinvent a fourth variant.

